PR #7, opened 2017-03-31
This is an alternative to pull request 4.

espindola:
It implements the same idea, but is a bit simpler IMHO.

### DIFF
--- a/lieer/remote.py
+++ b/lieer/remote.py
@@ -1,4 +1,5 @@
 import os
+import time
 import httplib2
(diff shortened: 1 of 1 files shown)


gauteh:
Hi,

This looks good. Please fix up the commit message and it should be good
to go. 

Perhaps we should have a max delay? Say 1000 seconds and just completely
fail on that, to prevent going forever.

gauteh:
#8

PR #8, opened 2017-03-31
Handle rate limit errors.

espindola:
If a rate limit is reported, add an exponential backoff before
retrying.

After 10 successful iterations reduce the delay again.


gauteh:
thanks, merged.

PR #17, opened 2017-05-02
remote: push: filter out duplicates not in this repository, and push …

gauteh:
…all duplicates in this repository.


gauteh:
@yeled: ping

yeled:
This lgtm. I don't know what else we could add - I'd merge with master.

gauteh:
Charlie Allom writes on mai 2, 2017 16:03:
> This lgtm. I don't know what else we could add - I'd merge with master.

Yes, I guess so. Only thing is to fix up the 'mid' to 'gid' where they
should be.

Issue #35, opened 2017-08-03
Sync SPAM and TRASH


Etn40ff:
At the moment  the value of remote.not_sync is fixed to 
`not_sync = set (['CHAT', 'SPAM', 'TRASH'])`
so that SPAM and TRASH are not downloaded.

In order to be able to undelete or remove from spam, I would need to sync these as well. Is it possible to make this a configurable option?

gauteh:
Check out #32, SPAM and TRASH is not currently possible to sync.

Etn40ff:
Sorry for the noise.
I am confused by your reply: if you just remove `TRASH` from `remote.not_sync` it seems to work (at least for new e-mails but not for stuff already in trash)
Can you point me to the right place in the API?

gauteh:
https://developers.google.com/gmail/api/v1/reference/users/messages/list

and i think you have to manually do a trash operation https://developers.google.com/gmail/api/v1/reference/users/messages/trash so it is not the same as just adding the label.

gauteh:
I dont remember the story behind the SPAM, TRASH stuff in the README, maybe @julian-klode can clarify?

julian-klode:
Well, you can add spam or trash labels locally, and the correct action will be done server-side on a push (that's how I mark emails as spam or delete them). But you can only add one of the inbox, spam, trash labels in a command - they are mutually exclusive.

That said, you then pull, and if you don't have includeSpamTrash set, you'll not see the emails anymore you just tagged spam or trash.

So I guess you can just make an "include spam & trash" toggle, make that toggle includeSpamTrash and the corresponding exclusion (whatever that does, I'm not sure - after all, it does push added spam and trash tags and that works as expected).

gauteh:
Ok.. not sure what remote.not_sync does then at the moment.

gauteh:
Ah right, so I've removed chat-messages by specifying so in the query. The CHAT messages are not really MIME messages and would typically fail to be synced in ways. I couldn't really read the content of them either.. a future feature might be to sync them and write out a mime message based on a template.

The trash and spam messages are not usually listed by messages.list, but by setting the flag to true the two TRASH/SPAM tags could be removed from the not_sync list. A full sync would be required after changing this. The behavior is a bit inconsistent now since we are pushing trash/spam changes, but not really pulling down existing trash/spam messages.

gauteh:
Please test #36 (carefully).

PR #36, opened 2017-08-06
remote: sync messages in TRASH or SPAM as well (#35)

gauteh:
Please test #35 

You need to do a full sync `gmi pull -f` to get these, consider trying out with dry-run first. Note that messages in GMail trash are deleted after 30 days, so if you add the trash/spam tag to local messages this may cause messages to be deleted.


Etn40ff:
This seems to work for me. 
Thanks

Issue #48, opened 2017-11-15
Missing GMail labels


ktf:
Similarly to #46 I've:

```
pull: full synchronization (no previous synchronization state)                                                                                    [3/8741]
fetching messages: 123866it [04:10, 414.05it/s]
receiving content:   0%|                                                                             | 3/57916 [00:00<9:22:58,  1.71it/s]Traceback (most r
ecent call last):
  File "/nix/store/5zb6zlyxg2xjz7p3kqr9hahl3vdbzm63-python3.6-gmaileer-v0.3-20171115/bin/.gmi-wrapped", line 9, in <module>
    g.main ()
  File "/nix/store/b0rgiilc0nhszygkgqy4sj47nhrbakzx-python3-3.6.2-env/lib/python3.6/site-packages/lieer/gmailieer.py", line 136, in main
    args.func (args)
  File "/nix/store/b0rgiilc0nhszygkgqy4sj47nhrbakzx-python3-3.6.2-env/lib/python3.6/site-packages/lieer/gmailieer.py", line 307, in pull
    self.full_pull ()
  File "/nix/store/b0rgiilc0nhszygkgqy4sj47nhrbakzx-python3-3.6.2-env/lib/python3.6/site-packages/lieer/gmailieer.py", line 530, in full_pull
    updated = self.get_content (message_gids)
  File "/nix/store/b0rgiilc0nhszygkgqy4sj47nhrbakzx-python3-3.6.2-env/lib/python3.6/site-packages/lieer/gmailieer.py", line 595, in get_content
    self.remote.get_messages (need_content, _got_msg, 'raw')
  File "/nix/store/b0rgiilc0nhszygkgqy4sj47nhrbakzx-python3-3.6.2-env/lib/python3.6/site-packages/lieer/remote.py", line 109, in func_wrap
    return func (self, *args, **kwargs)
  File "/nix/store/b0rgiilc0nhszygkgqy4sj47nhrbakzx-python3-3.6.2-env/lib/python3.6/site-packages/lieer/remote.py", line 282, in get_messages
    batch.execute (http = self.http)
  File "/nix/store/b0rgiilc0nhszygkgqy4sj47nhrbakzx-python3-3.6.2-env/lib/python3.6/site-packages/oauth2client/util.py", line 137, in positional_wrapper
    return wrapped(*args, **kwargs)
[...]

PR #49, opened 2017-11-16
Workaround #48

ktf:
This adds a notmuch "unknown" tag whenever a label id which cannot be found
in the list of available label ids is retrieved from a message. Notice the
"unknown" tag is never synchronized.

### DIFF
--- a/lieer/local.py
+++ b/lieer/local.py
@@ -35,6 +35,7 @@ class Local:
                         'replied',
                         'muted',
                         'mute',
+                        'unknown'
                         ])
 
   class RepositoryException (Exception):
@@ -351,9 +352,10 @@ class Local:
     labels = m.get('labelIds', [])
 
     # translate labels. Remote.get_labels () must have been called first
-    labels = [self.gmailieer.remote.labels[l] for l in labels]
-
-    labels = set(labels)
+    newLabels = []
+    for l in labels:
+        newLabels.append(self.gmailieer.remote.labels.get(l, "unknown"))
+    labels = set(newLabels)
 
     # remove ignored labels
     labels = list(labels - self.gmailieer.remote.ignore_labels)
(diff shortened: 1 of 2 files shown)


gauteh:
Thanks; good that you figured out a way to get around this. I don't think I want to merge this right away since there must be some other underlying error. Could you try to put in the offending label in https://developers.google.com/gmail/api/v1/reference/users/labels/get ? Maybe we have to manually fetch the label if it didn't get fetched the first time.

gauteh:
If we can confirm that this is indeed a non-existant label, maybe it would be better to just not add anything and ignore it. It would possibly delete the label on the GMail side in the future so I think we should have a switch which allows this. It is likely to only be an issue on initial sync anyway. Then we can just advice the user to enable this switch if he wants.

ktf:
Actually, the label is not there at all, neither in list, nor if I get it by id:

```
{
 "error": {
  "errors": [
   {
    "domain": "global",
    "reason": "notFound",
    "message": "Not Found"
   }
  ],
  "code": 404,
  "message": "Not Found"
 }
}
```

other labels are actually there.

ktf:
And if I look into the message itself, the label is actually there:

```
 "labelIds": [
  "Label_3",
  "CATEGORY_PERSONAL"
 ],
```

gauteh:
Giulio Eulisse writes on november 17, 2017 0:18:
> And if I look into the message itself, the label is actually there:
> 
> ```
>  "labelIds": [
>   "Label_3",
>   "CATEGORY_PERSONAL"
>  ],
> ```

Hm, weird! This looks like a GMail bug. Do you want to report the issue 
there: https://issuetracker.google.com/ ?

Not much hope in getting an answer, I already have one pending:

https://issuetracker.google.com/issues/62686910

ktf:
I will try. I am also trying to see if another route works better.

gauteh:
Check out the approach in #51, does that work for you?

gauteh:
Implemented through #51.

ktf:
thanks. That's what I needed.

Issue #40, opened 2017-09-09
Speed up initial download


manishrjain:
My personal email goes back over 10 years, and I'd rather not have all of that downloaded on my laptop (even setting the max messages per label in Gmail doesn't help with that, because I have gained way too many labels over the years). offlineimap allows you to sync mails from certain 'folders'. Would be good if gmailieer allows that.

gauteh:
The mail that should be synced could perhaps be limited by speifying a 
query (https://github.com/gauteh/gmailieer/blob/master/lieer/remote.py#L64). 
The problem is that e.g. history().list(..) does not take a query, so 
things would get messy when doing a partial update. There are probably 
other methods that should be limited to the query as well.

Note that once you have all your emails synced the partial update will 
be fast. Is this in order to save disk space?

-g

Manish R Jain writes on september 9, 2017 8:21:

manishrjain:
Disk space isn't that much of a concern (might be, but I don't know because I haven't yet done a full sync). But, just the process of downloading a decade of emails on your laptop is really slow, and very resource heavy. Also, seems a bit pointless, because when on laptop, you just want to get your inbox to zero, not much else. Also, offlineimap does this out of the box.

I can see why partial sync would be tricky -- IIUC, it would require keeping an external state of all the messages that you had previously downloaded but don't belong to any of the labels in sync config anymore. So, the right question is, does the time it takes to get up and running with Gmailieer and Astroid on a laptop outweigh the complexity of doing partial syncs? I'd say yes.

gauteh:
Manish R Jain writes on september 11, 2017 2:59:
> Disk space isn't that much of a concern (might be, but I don't know because I haven't yet done a full sync). But, just the process of downloading a decade of emails on your laptop is really slow, and very resource heavy. Also, seems a bit pointless, because when on laptop, you just want to get your inbox to zero, not much else. Also, offlineimap does this out of the box.
>
> I can see why partial sync would be tricky -- IIUC, it would require keeping an external state of all the messages that you had previously downloaded but don't belong to any of the labels in sync config anymore. So, the right question is, does the time it takes to get up and running with Gmailieer and Astroid on a laptop outweigh the complexity of doing partial syncs? I'd say yes.
>

When I said partial sync, incremental synchronization might have been
more accurate. This is what happens automatically if you have a recent
enough state file (have done partial pulls with not too long intervals).

For me, keeping my entire mail archive locally, indexed by notmuch and
[...]

PR #55, opened 2017-11-28
remote: fetch messages using IMAP

gauteh:
__Not very well tested!__

Use IMAP to fetch message content.

This requires:
* fetch message list
* for each message
  - fetch IMAP UID for GMail msgid
  - fetch message content
* fetch metadata for all messages in batch requests (to get labels)

Requires extended scope (access rights). Try with `gmi pull --imap`.

### DIFF
--- a/lieer/gimap.py
+++ b/lieer/gimap.py
@@ -1,0 +1,73 @@
+import imaplib
+import base64
+
+class Gimap:
+  """
+  * https://github.com/google/gmail-oauth2-tools/blob/master/python/oauth2.py#L255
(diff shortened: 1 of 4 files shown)


yeled:
why? :)

gauteh:
Was an attempt at speeding up initial download (#40) - but didn't improve 
speed! At the moment I don't fetch messages in batches (using IMAP), but 
one and one - so that might be one potential slow-down.

Charlie Allom writes on desember 13, 2017 11:10:

gauteh:
Closing this. Slower than API.

PR #61, opened 2018-01-30
remote: index by batches

gauteh:


lieer/gmailieer.py
```diff
@@ -593,11 +593,12 @@ def get_content (self, msgids):
 
       bar = tqdm (leave = True, total = len(need_content), desc = 'receiving content')
 
-      def _got_msg (m):
-        bar.update (1)
-        # opening db per message since it takes some time to download each one
+      def _got_msg (ms):
```
julian-klode: Maybe call it `_got_msgs` instead to make it more clear that its multiple ones.

gauteh: All testing appreciated.

PR #62, opened 2018-02-15
Label to Tag translation  & Tag filtering

amitramon:
This pull request adds a new feature to gmailieer - the user can now define how GMail labels are translated to Notmuch tags when syncing Gmail and Notmuch. In addition, the user can set filters that define which local Notmuch tags are excluded from syncing to GMail. Also, the user can define a local label separator that will replace the GMail label separator when syncing to Notmuch.

Detailed information about the change can be found in the docstring for the new module label_translation.py and in the updated README file.

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -13,14 +13,54 @@ from tqdm import tqdm, tqdm_gui
 
 from .remote import *
 from .local  import *
+from .labels_translation import LabelTranslator
 
 class Gmailieer:
+
+  # the name of the custom user's label translation file. can be
+  # either relative or full path.
+  user_label_trans_file_name = '.label-trans.json'
+
   def __init__ (self):
     xdg_data_home = os.getenv ('XDG_DATA_HOME', os.path.expanduser ('~/.local/share'))
     self.home = os.path.join (xdg_data_home, 'gmailieer')
 
+    # the sole instance of LableTranslator
+    self._label_translator = LabelTranslator()
+
+  @property
+  def label_translator(self):
+    """Return the instance of LabelTranslator"""
+    return self._label_translator
+
+  def show_label_translation(self, args):
+    """Show label translation map and additional translation information"""
+    if args.default_map:
+      # no map file, default translation
+      LabelTranslator.print_default_translation()
+    elif args.map_from_file:
+      # tranlation from arbitrary user supplied file
+      lt = LabelTranslator()
+      try:
+        lt.load_user_translation(args.map_from_file)
+        print("User's label translation loaded (file: {})".format(args.map_from_file))
+      # except json.decoder.JSONDecodeError as e:
+      except Exception as e:
+        print('Failed to load user map file: {}'.format(args.map_from_file))
+        print(e)
+        raise
+      lt.print_info()
+    else:
+      # print _actual_ translation, based on state settings
+      self.setup(args, dry_run=False, load=True)
+      self.label_translator.print_info()
+
   def main (self):
-    parser = argparse.ArgumentParser ('Gmailieer', parents = [tools.argparser])
+    parser = argparse.ArgumentParser(prog='gmi',
+                                     description="Sync email between GMail and Notmuch database",
+                                     epilog="To get help on a specific command use the following:"
+                                     " %(prog)s <command> -h",
+                                     parents=[tools.argparser])
     self.parser = parser
 
     common = argparse.ArgumentParser (add_help = False)
@@ -105,13 +145,18 @@ class Gmailieer:
         description = 'initialize',
         help = 'initialize local e-mail repository and authorize')
 
-    parser_init.add_argument ('--replace-slash-with-dot', action = 'store_true', default = False,
-        help = 'This will replace \'/\' with \'.\' in gmail labels (make sure you realize the implications)')
-
     parser_init.add_argument ('--no-auth', action = 'store_true', default = False,
         help = 'Do not immediately authorize as well (you will need to run \'auth\' afterwards)')
 
     parser_init.add_argument ('account', type = str, help = 'GMail account to use')
+
+    # Enabel user labels translation
+    user_label_translatien_help = 'Translate labels according to user-supplied map. ' \
+                                  + 'Read the documentation and make sure you understand ' \
+                                  + 'the implications'
+
+    parser_init.add_argument('--user-label-translation', action='store_true',
+                             help=user_label_translatien_help)
 
     parser_init.set_defaults (func = self.initialize)
 
@@ -124,27 +169,43 @@ class Gmailieer:
     parser_set.add_argument ('-t', '--timeout', type = float,
         default = None, help = 'Set HTTP timeout in seconds (0 means system timeout)')
 
-    parser_set.add_argument ('--replace-slash-with-dot', action = 'store_true', default = False,
-        help = 'This will replace \'/\' with \'.\' in gmail labels (make sure you realize the implications)')
-
-    parser_set.add_argument ('--no-replace-slash-with-dot', action = 'store_true', default = False)
-
     parser_set.add_argument ('--drop-non-existing-labels', action = 'store_true', default = False,
         help = 'Allow missing labels on the GMail side to be dropped (see https://github.com/gauteh/gmailieer/issues/48)')
 
     parser_set.add_argument ('--no-drop-non-existing-labels', action = 'store_true', default = False)
 
+    # Enabel or disable user label translation
+    group_set = parser_set.add_mutually_exclusive_group()
+    group_set.add_argument("--user-label-translation", action="store_true",
+                       help=user_label_translatien_help)
+    group_set.add_argument("--no-user-label-translation", action="store_true",
+                       help='Disable user\'s label translation')
+
     parser_set.set_defaults (func = self.set)
 
+    # show label tranlation detailed information
+    parser_label_trans = subparsers.add_parser ('show-label-translation', parents=[common],
+        description = 'Show label translation',
+        help = 'show the current label translation map which would be used for sync')
 
+    group_label_trans = parser_label_trans.add_mutually_exclusive_group()
+    group_label_trans.add_argument ('-d', '--default-map', action='store_true', default=False,
+                              help='show only the default map')
+    group_label_trans.add_argument ('-f', '--map-from-file',
+                              help='show the map that would be created for a given map file')
+
+    parser_label_trans.set_defaults(func=self.show_label_translation)
+
+    # run the selected command
     args        = parser.parse_args (sys.argv[1:])
     self.args   = args
-
     args.func (args)
+
 
   def initialize (self, args):
     self.setup (args, False)
-    self.local.initialize_repository (args.replace_slash_with_dot, args.account)
+    self.local.initialize_repository(args.account,
+                                     args.user_label_translation)
 
     if not args.no_auth:
       self.local.load_repository ()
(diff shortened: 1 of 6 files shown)


gauteh:
Thanks! Quite big changes, so wont be able to properly review before next week. If someone else tests, please note here as well the results. +1 for making `replace-slash-with-dots` more general.

Will there be any changes to the current behavior if no custom configurations is made? This is quite important since it could cause loads of changes to be propagated. You mention that things could get out of sync; would a way around this be to do `gmi push -f` after having manually changed the local tags to match the new scheme. Or perhaps `gmi pull -f` is better?

If a user currently has replace-slash-with-dots set; we should print an error and exit until a matching custom configuration has been set / or the key has been removed from the current config file - so that the behavior won't change without the user knowing.

lieer/labels_translation.py
```diff
@@ -0,0 +1,332 @@
...
+        'todo',
+        'trash',
+        'bin',
+        'archive'
+    }
+
+    # Note: some tests against Gmail (done by amitramon) show that not
+    # all of the labels above are rejected by Gmail.
+    #
+    # 'attachment', 'encrypted', 'signed', 'passed' and 'replied' seem
```
gauteh: These were ignored because they are added automatically by notmuch: https://notmuchmail.org/special-tags/

lieer/labels_translation.py
```diff
@@ -0,0 +1,332 @@
...
+    # this is the reason for its rejection.
+    #
+    # 'new' seems not to be a Gmail reserved label, but I guess is
+    # here because of its part in the 'standard' Notmuch workflow. If
+    # that is indeed the case, it would be better to leave the
+    # decision whether to ignore it or not to the end user.
+    #
+    # Last but not least, 'trash' is a special case. Gmail accepts
+    # 'TRASH' in all upper-case letters, but rejects it in lower or
+    # mixed case. This is handled by the code bellow as a special case.
```
gauteh: Do you take into account that only one of spam, inbox and trash can be applied at the same time? See last section in readme.

lieer/labels_translation.py
```diff
@@ -0,0 +1,332 @@
...
+        'CHAT'      : 'chat'
+    }
+
+    # set of Gmail's system labels that should not be created by gmi.
+    # the labels in this set should be written in all lower case letters.
+    SYSTEM_IGNORED_LABELS = {
+        'attachment',
+        'encrypted',
+        'signed',
+        # 'new',
```
gauteh: I think new is set as new-tag for new notmuch configs. So I think it should also be ignored by default, but that should be configurable along with the rest of the custom tag-ignoring.

amitramon: According to the notmuch-config man page (on Debian), the default tags that are assigned to new mail are 'unread' and 'inbox'. It seems, however, that tagging new mail with 'new' is a common convention.

But since the choice of the tag(s) for new mail are dependant on conventions or user preferenc, in my opinion gmailieer should not hard-code it, and should let the user to decide. So my suggestion is to remove 'new' from the SYSTEM_IGNORED_LABLES list, and in case the user decide to use 'new' for tagging new mail, he or she could use the new filter mechanism to exclude this tag from syncing to GMail.

lieer/labels_translation.py
amitramon: Label Translation doesn't change any existing logic - it merely maintains a 1-1 map between local and remote tags. So the trash/spam/inbox mechanism was not changed.

The only thing that can be possibly be different is that the user can use 'custom' names for the local tags - for example, 'junk' instead of 'spam' and 'deleted' instead of 'trash'. But this should not affect the logic - the check of these three labels in the update_tags method is done after the tags are translated to the remote system, and it should be okay.

lieer/labels_translation.py
amitramon: You're right, I missed that. So (explaining to myself), some of the ignored labels are Notmuch-internals and thus should not be synced to GMail, and some are labels that GMail reserved labels that are not allowed to be created.

Bottom line, it's all okay, and my long comment in the code could be removed or modified.

amitramon:
Regarding your first comment:

> Will there be any changes to the current behavior if no custom configurations is made? This is quite important since it could cause loads of changes to be propagated.

No, if no custom configuratin is made the system's behavior will be exactly the same as before this feature, with just one exception - I commented out the 'new' tag in the list of ignore-labels. So now 'new' will, by default, be synced to GMail. But this is the only change.

> You mention that things could get out of sync; would a way around this be to do gmi push -f after having manually changed the local tags to match the new scheme. Or perhaps gmi pull -f is better?

This is a good point, I'm not sure I have an answer right now. If I understand your suggestion, it could be a solution, although, maybe, expnsive and perhaps risky - it would require, for example, to remove all the remote lables on all messages, then assign them the local tags (or vice versa). The best thing would be to be able not to get to such situations. I'll try to think about it.

> If a user currently has replace-slash-with-dots set; we should print an error and exit until a matching custom configuration has been set / or the key has been removed from the current config file - so that the behavior won't change without the user knowing.

This is another good point, and, in general, we should think about migration from the current configuration to the new one with custom translation/label-separator. But the minimum should be like you said - the system should be able to detect that things have changed and alert the user. Another thing for me to think about.

lieer/labels_translation.py
gauteh: Ok, sounds good. Lets keep 'new' outside the synced tags - but we then need to notify users who don't add it to their new label ignore list.

lieer/gmailieer.py
```diff
@@ -617,25 +691,57 @@ def set (self, args):
...
+    reply = str(input(question + ' [y/N]: ')).lower().strip()
+
+    if reply[:1] == "y":
+      self.local.state.set_user_label_translation(new_label_translation_value)
+      print(">> User label translation {}".format(
+        ['DISABLED', 'ENABLED'][new_label_translation_value]))
+    else:
+      print(">> User label translation setting not changed")
+
+    print("")
```
gauteh: I would like to keep interactiveness at a minimum (we have to be interactive with getting oauth tokens). I think it would be better to use something like:

```
* gmi set --user-label-translation file -> only disaplays warning
* gmi set --user-label-translation --yes file actual sets option
```

Or we could just have no warning on setting but list it in (partially) big types in the help output / readme.

lieer/labels_translation.py
```diff
@@ -0,0 +1,332 @@
...
+        'passed',
+        'replied',
+        'muted',
+        'mute',
+        'todo',
+        'trash',
+        'bin',
```
gauteh: `bin` needs to be added to the README caveats section.

lieer/remote.py
```diff
@@ -656,6 +644,15 @@ def __create_label__ (self, l):
         if excep.resp.status == 403 or excep.resp.status == 500:
           self.__request_done__ (False)
           return self.__create_label__ (l)
+        # not sure this is good enough, but it would be nice to be
+        # able to notify the end user that some specific tags were
+        # rejected by Gmail and then continue, without exiting with an
+        # exception.
```
gauteh: 👍

lieer/remote.py
```diff
@@ -656,6 +644,15 @@ def __create_label__ (self, l):
...
+
+        # elif excep.resp.status == 400:
+        #   print("REJECTION: Label '{}' rejected by Gmail".format(l))
+        #   return (None, None)
```
gauteh: Unless we have a solution we should probably still do `raise` in the end

gauteh:
Amit Ramon writes on februar 15, 2018 17:45:
>> You mention that things could get out of sync; would a way around this be to do gmi push -f after having manually changed the local tags to match the new scheme. Or perhaps gmi pull -f is better?
> 
> This is a good point, I'm not sure I have an answer right now. If I understand your suggestion, it could be a solution, although, maybe, expnsive and perhaps risky - it would require, for example, to remove all the remote lables on all messages, then assign them the local tags (or vice versa). The best thing would be to be able not to get to such situations. I'll try to think about it.
> 

Yeah - remote-to-local would probably be faster if lots of messages need 
to be changed. A full sync is necessary, but it would only be metadata, 
not the content so not as heavy as the initial sync. Writing a migration 
handler would probably be too hard to get right. If the translation 
rules are changed, but there shouldn't be an effect on the remote, then 
the fastest is probably to manually change the local tags to match the 
new scheme and then do a push. If you're sure that no change should be 
pushed you could just manually set the lastid in the repostiory file to 
the current revision so that gmi things taht all local changes have been 
pushed.

lieer/gmailieer.py
amitramon: just to make sure - 'gmi set --user-label-translation' sets a boolean flag in the state file. 
Just to verify I understand your suggestion, do you mean that without the 'force' switch we should issue a warning similar to the current one, and without the 'force' switch we should go ahead and set the flag without warnings?
If this is it, I think it would be a good solution.

lieer/labels_translation.py
amitramon: I agree. (and `archive` too)

lieer/remote.py
amitramon: I think we need to assume that we don't have exact information as for what are all the GMail's reserved labels (that, at least, is the impression I have) and also to assume that this can be changed in the future.
This is why I think that when a user's tag is rejected by GMail, we should just issue a warning, so the user can change the tag (or define a translation for it).
An exeption is surely a kind of warning, but I believe that end users should not get a exception in this case, and that a warning or an error message would be enough. But I'll the decision to you.
On the technical side, what I tried was to return a value that would let the program go on. (None, None) was causing errors, so perhaps you would know what is the right solution here, if we go that way.

amitramon:
Actually after making some tests and thinking about it, there is some 'big' problem. What I had in my mind while writing this thing was that it will be used for gmi for the very first time, without having mail messages and tags created by a previous version.

Now the situation when messages and tags already exist is different. Just as an example, lets assume we have a remote label `remote` that, before this version, was related to a local tag of the same name, Suppose we now want to translate it to a tag called `local`, so we go ahead and define the mapping in the `.label-trans.json` file: `remote <--> local`.

But what we actually have, is a local tag `remote` and a remote label `remote`. if we push, nothing will change. If we pull, I was expecting that `pull` or `pull -f` will change the local tag, but it doesn't seem to work. I'm not sure whether this is the correct operation of pull or not, and I still haven't got the time to debug it.

The bottom line is that for an initial synchronization I think this is a good feature. For a change from a previous version, either we should improve it, or - maybe not the best option - have the user to sync labels-tags by hand before running the new version with a label translation map.

This discussion also holds for the label separator issue, since it is just a (maybe slightly different) translation.

This is, in my opinion, the major issue that we need to consider.

gauteh:
I agree. Full downloads are too big, so we cannot expect our users to do that as a solution to a new config.

amitramon:
Hello, sorry for the late reply, had a busy week. What I'd like to suggest is that I'll go over all the discussion here again, and try to get a coherent view as for what changes are needed and what I can do.
I'll then update the pull request, so we'll have a new basis to refer to.
Unfortunately, due to current pressure, it'll take me a week or a week and a half to get to it.

Best, Amit

gauteh:
Amit Ramon writes on februar 25, 2018 16:38:
> Hello, sorry for the late reply, had a busy week. What I'd like to suggest is that I'll go over all the discussion here again, and try to get a coherent view as for what changes are needed and what I can do.
> I'll then update the pull request, so we'll have a new basis to refer to.
> Unfortunately, due to current pressure, it'll take me a week or a week and a half to get to it.

No problem. When you do get to this it would make it easier to merge and 
review if you split your changes into several smaller commits, preferably those 
that do not cause any change in behaviour first. Then I can merge those 
immediately and we can keep improving the rest untill we have a working 
migration process. Migration does complicate things a bit here I agree.

gauteh:
By the way; label-separator seems like a fairly common special case of translations, but would it make sense to just incorporate that into the regular translations?

dtzWill:
As someone rather interested in this functionality, friendly ping! :D

gauteh:
Closing because of inactivity. This functionality would be great, but it is involved to get right.

Issue #75, opened 2018-04-24
Updating tags on non-existent file


stdbull:
After performing a sync of my pretty large mailbox, when updating tags, I consistently see an exception about failing to update tags on a non-existent file. I think gmailieer may have somehow got the file name wrong, but I'm not sure... Here's the trace:
...
resolving changes: 100%
receiving content: 100%
updating tags (0Δ):   0%|                                                                                   | 0/13602 [00:00<?, ?it/s]Traceback (most recent call last):
  File "/usr/local/bin/gmi", line 8, in <module>
    g.main ()
  File "/b/git/gmailieer/lieer/gmailieer.py", line 145, in main
    args.func (args)
  File "/b/git/gmailieer/lieer/gmailieer.py", line 215, in sync
    self.pull (args, True)
  File "/b/git/gmailieer/lieer/gmailieer.py", line 341, in pull
    self.partial_pull ()
  File "/b/git/gmailieer/lieer/gmailieer.py", line 492, in partial_pull
    r = self.local.update_tags (m, None, db)
  File "/b/git/gmailieer/lieer/local.py", line 401, in update_tags
    raise Local.RepositoryException ("tried to update tags on non-existant file: %s" % fname)
lieer.local.RepositoryException: tried to update tags on non-existant file: /path/to/toplevelmaildir/mail/cur/162f6b5441a3ab90:2,

The file "/path/to/toplevelmaildir/mail/cur/162f6b5441a3ab90:2," does not exist in the mail store, but file "/path/to/toplevelmaildir/mail/cur/162f6b5441a3ab90:2,S" does.

Could it be possible that gmailieer has had a problem parsing this file name?

notmuch search for "/path/to/toplevelmaildir/mail/cur/162f6b5441a3ab90:2,S" correctly identifies the email and shows the tags applied to it. The message only has one mime part which is text/html (no text/plain part).

Also, is there a particular reason why this exception hangs gmailieer? I think it would be better if it exited.
--  Matt

gauteh:
Are you by any chance reading your email while gmi is syncing? It seems 
that the 'S' (seen) flag is added to a message. This will happen if the 
message has been read by a MUA and the maildir flags have been synced. 
No changes to the messages should be made while gmi is syncing.

I don't see how it can hang.. it should either exit or just 
continuing to sync the next messages.

Matthew Lear writes on april 24, 2018 11:40:

stdbull:
[...]

PR #77, opened 2018-04-26
Singleton

gauteh:
#75


julian-klode:
A socket is crazy, why not use `fcntl.lockf()` or just create a file with O_EXCL and delete it when program ends.

julian-klode:
or `fcntl.flock()`

julian-klode:
Apparently lockf is the most portable option. And they are bound to open file descriptors somehow (so closing the fd would release the lock). I can't remember which lock is which, but some are inheritable by subprocess and some not, some are bound to files instead of file descriptors (so you lock one fd, if you close another fd refering to the same file in the same process, the lock goes away)I don't think you have subprocesses or multiple fds for the same file - especially if you use a new lock file, so nothing to worry about.

gauteh:
Can you test?

PR #92, opened 2018-09-27
readme: add caveat that excluded_tags are not honored

### LINKED ISSUE #91: pushing wrong maildir to Gmail
I woke up to a nasty surprise this morning: my work Gmail had thousands of new messages uploaded to it that don't belong there. I don't know if this is user error on my part or if there is a bug in how gmailieer chooses which messages to upload. Here is my configuration:

`~/.mail/work` is the gmailieer working directory for my main account. This has been working fine for a while. I have a cron job that pulls, does some tagging, and then pushes via `gmi`.

I have just added `~/.mail/archive_foo` and `~/.mail/archive_bar` that are both maildirs of dead accounts. They come from offlineimap days, before gmailieer, and I do not sync or otherwise change anything about these maildir. I do want them indexed by notmuch. They are just for historical searches.

When I initially indexed each archive I tagged all of the mails: `notmuch tag +archived-foo path:/foo/`, etc.

This morning my Gmail web interface shows a new Label, `archived-foo` with 28k emails in it :-(

Any ideas on why this happened? Is it expected behavior? For reference my cronjob, simplified for readability:

```
#!/usr/bin/env fish

set PATH ~/.config/alot/hooks $PATH

cd ~/.mail/work/

gmi pull

# fetch patwork patch status and convert into notmuch tags
fetch-pw-states-update-notmuch-tags.fish

gmi push

notmuch tag -new -- tag:new
```

So the surprising thing to me is that this script resulted in messages from outside of `~/.mail/work` getting uploaded to the gmail account associated with `work`. Is this expected behavior?

More generally, does Gmailieer's design play nice with multiple Google accounts indexed into a single notmuch database? I have to assume that a common pattern amongst Gmaillieer users is to have a work gmail and a personal gmail account, indexed together.

mturquette:
Closes #91

Signed-off-by: Michael Turquette <mturquette@deferred.io>

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,3 +189,5 @@ the time being, `trash` will be prefered over `spam`, and `spam` over inbox.
 
 * Make sure that you use the same domain for you GMail account as you initially created your account with: usually `@gmail.com`, but sometimes `@googlemail.com`. Otherwise you might get a [`Delegation denied` error](https://github.com/gauteh/gmailieer/issues/88).
 
+* Any tags listed as `excluded_tags` in `.notmuch-config` will still be synchronized by Gmailieer. Use `--ignore-tags-local` or `--ignore-tags-remote` as appropriate to prevent synchronizing tags between the local notmuch repository and GMail.
+


gauteh:
Thanks, maybe make it a bit more general?

Only the `new.tags` option is read from `.notmuch-config`. Options 
like `exclude_tags` are therefore not respected (see `gmi set` for similar 
functionality).

Maybe link to the issue around `not respected` and link to the section 
in the README with `gmi set`. Note the typo on `exclude_tags`.

Michael Turquette writes on September 27, 2018 23:01:

gauteh:
Closing.

Issue #98, opened 2019-01-04
suggest: add option to specify file extension


tbrk:
I am the author of [tbrk/muttlight](https://github.com/tbrk/muttlight) which improves the indexing of messages in maildirs on MacOS using the "Spotlight" feature.

This software must be registered with the OS to treat the files stored in maildirs. The simplest way to associate files on MacOS with the applications that treat them, i.e., for opening or indexing, is to register file extensions. The fetchmail and getmail programs create maildir files with the extension `.mbox:2,DFPRST` or `.<hostname>:2,DFPRST` for all combinations of the maildir status flags `DFPRST` and where the current `<hostname>` is sometimes used to facilitate synchronization of maildir folders across different machines.

I don't use gmailieer myself, but I have been contacted by someone who does asking whether he could also use muttlight. Unfortunately, I do not know how to do this reliably without using file extensions.

So, would it please be possible to add a configuration option whereby gmailieer would add a period and the current hostname before the maildir version and status flags in generated filenames?

gauteh:
Hi,

gmailieer cannot be used without notmuch, if still useful adding a configurable portion would be possible. Since we already have to remove the last portion of the filename this should not have any additional performance effect. The maildir flags are not static though, would that not change the extension?

Regards, Gaute

tbrk:
Yes, I think it would be useful. I use mutt with [mu](https://www.djcbsoftware.nl/code/mu/) on MacOS and having Spotlight indexing and preview is still very useful in the GUI. I expect the same to be true for notmuch on MacOS.

The maildir flags are indeed problematic. In fact, they are one of the problems that muttlight tries to solve. The solution is not especially clever: we simply register all 64 combinations of the flags for all user-specified hostnames. It seems to work quite well in general.

gauteh:
Ok, this should not be difficult. I dont have the chance to do this in the next days, but happy to review a PR if someone wants to give it a shot.

tbrk:
I've had a shot at adding this feature. I've done some testing, but I'm a complete gmailieer/notmuch newbie so my patch probably needs a careful review.

I've assumed that synchronized filenames never usually contain a period ("."). Is this reasonable?

gauteh:
Fixed in #99

PR #99, opened 2019-01-04
Option to store files with a given extension

tbrk:
Adding file extensions aims to help MacOS users index their synchronized
gmail messages with Spotlight (using https://github.com/tbrk/muttlight).

To use:
    gmi set --file-extension mbox

/!\ This patch assumes that filenames contain a period only if a file
extension has been added using the option.

Addresses #98

### DIFF
--- a/lieer/local.py
+++ b/lieer/local.py
@@ -66,6 +66,7 @@ class Local:
     drop_non_existing_label = False
     ignore_tags = None
     ignore_remote_labels = None
+    file_extension = None
 
     def __init__ (self, state_f):
       self.state_f = state_f
@@ -84,6 +85,7 @@ class Local:
       self.drop_non_existing_label = self.json.get ('drop_non_existing_label', False)
       self.ignore_tags = set(self.json.get ('ignore_tags', []))
       self.ignore_remote_labels = set(self.json.get ('ignore_remote_labels', Remote.DEFAULT_IGNORE_LABELS))
+      self.file_extension = self.json.get ('file_extension', '')
 
     def write (self):
       self.json = {}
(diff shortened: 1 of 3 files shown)


tbrk:
I've had a shot at adding this feature. I've done some testing, but I'm a complete gmailieer/notmuch newbie so my patch probably needs a careful review.

I've assumed that synchronized filenames never usually contain a period ("."). Is this reasonable?

lieer/local.py
```diff
@@ -303,16 +310,23 @@ def messages_to_gids (self, msgs):
...
-          gid = os.path.basename (fname).split (':')[0]
+          gid = self.__filename_to_gid__ (os.path.basename (fname))
           gids.append (gid)
           messages.append (m)
 
     return (messages, gids)
 
 
+  def __filename_to_gid__ (self, fname):
+    return fname.split (':')[0].split('.')[0]
```
gauteh: I think it would be better to use the full configured extension, perhaps using `rfind`, presumably `:` can be added as part of the search key to avoid two splits/rfinds and make the splitting unambigious. This will fail now if the configured extension contains `.` or `:`. We also need to add a section to the README about changing this option after synchronization similar to a few of the other options. Essentially you need to delete existing files and re-sync everything unless you manually rename all the files (since we are not providing a migration).

gauteh:
If it weren´t for the fact that it would cause a massive migration headache I wouldn´t mind having a default value for the file extension.

lieer/local.py
tbrk: I thought about using the full extension, but there would be problems if that extension was changed without renaming mail synchronized with a previous extension.

I agree though that the `--file-extension` option should reject arguments that contain `.` or `:`. I hadn't thought of that.

My idea was that migration is a separate problem, solved either with a full resync as you suggest, or with a simple shell script.

tbrk:
Yes, I was worried that adding a default value may upset existing users.

lieer/local.py
tbrk: OK. You should, of course, edit my patch as you see fit.

gauteh:
Do you have a test set up? See the following change: d677db768e5a0162b61a51ce5ace03a799c36edb

gauteh:
Merged with my changes.

PR #105, opened 2019-04-15
Separate configuration and state

### LINKED ISSUE #101: Separate synchronization state from configuration
I use [nix](https://nixos.org/nix/) to manage all of my software configurations. What it does is generate immutable configuration files in a store, then symlinks those files to the correct locations for software that cannot be configured to look in arbitrary locations for config files.

I have a problem doing this for Gmailieer, because Gmailieer modifies `.gmailieer.json` after each pull with ephemeral state:

- `last_historyId`
- `lastmod`

Perhaps these could be stored in a separate file, or an "immutable" configuration could be provided?

An alternative is to wrap `gmi` with CLI configuration options, but I'd rather generate config files.

NickHu:
Resolves #101 

This change introduces a configuration file, stored at `$XDG_CONFIG_HOME/gmailieer/config` (which is `~/.config/gmailieer/config` on most Linux systems) which contains the sync settings. Multiple accounts utilise the same configuration file, with their associated credentials still in the same places as previously (e.g.\ `~/.mail/personal/.credentials.gmailieer.json`, `~/.mail/work/.credentials.gmailieer.json`).

In addition, `gmi` commands are callable irrespective of their current directory, and all subcommands now expect an account to operate on (e.g. `cd ~/.mail/personal && gmi sync` becomes `gmi sync my.email@gmail.com`, where you would have something like
```ini
[my.email@gmail.com]
dir=personal
```
in your config).

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -29,6 +29,8 @@ class Gmailieer:
     common.add_argument ('-s', '--no-progress', action = 'store_true',
         default = False, help = 'Disable progressbar (always off when output is not TTY)')
 
+    common.add_argument ('account', type = str, help = 'GMail account to use')
+
     subparsers = parser.add_subparsers (help = 'actions', dest = 'action')
     subparsers.required = True
 
@@ -112,8 +114,6 @@ class Gmailieer:
 
     parser_init.add_argument ('--no-auth', action = 'store_true', default = False,
         help = 'Do not immediately authorize as well (you will need to run \'auth\' afterwards)')
-
-    parser_init.add_argument ('account', type = str, help = 'GMail account to use')
 
     parser_init.set_defaults (func = self.initialize)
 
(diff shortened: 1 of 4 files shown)


lieer/local.py
```diff
@@ -115,52 +97,39 @@ def set_lastmod (self, m):
...
-        self.write ()
-      except OSError:
-        print ("Failed creating test file with file extension: " + t + ", not set.")
-        raise
-
   def __init__ (self, g):
     self.gmailieer = g
-    self.wd = os.getcwd ()
+
+    xdg_config_home = os.getenv ('XDG_CONFIG_HOME', os.path.expanduser ('~/.config'))
```
tadfisher: Thank you for this.

lieer/gmailieer.py
```diff
@@ -29,6 +29,8 @@ def main (self):
     common.add_argument ('-s', '--no-progress', action = 'store_true',
         default = False, help = 'Disable progressbar (always off when output is not TTY)')
 
+    common.add_argument ('account', type = str, help = 'GMail account to use')
```
tadfisher: Hmm, would it be feasible to allow an empty value here mean "all accounts"?

lieer/local.py
```diff
@@ -259,9 +228,10 @@ def initialize_repository (self, replace_slash_with_dot, account):
       raise Local.RepositoryException ("'mail' exists: this repository seems to already be set up!")
 
     self.state = Local.State (self.state_f)
-    self.state.replace_slash_with_dot = replace_slash_with_dot
-    self.state.account = account
-    self.state.write ()
+    self.state.write()
+    self.config.set(self.gmailieer.args.account, "replace_slash_with_dot", "true" if replace_slash_with_dot else "false")
+    with open(self.config_f, mode="w") as fd:
+      self.config.write(fd)
```
tadfisher: How would this work with a non-writable config file? If I'm reading this right, any time the user specifies these values in the root config, they're copied to the account section and immediately written. This wouldn't exactly work if the config file can't be written.

NickHu: You're right, this needs to be fixed - I'll do it soon

lieer/gmailieer.py
NickHu: I'm not sure that's so elegant or intuitive. Perhaps we should have a default account in the global section of the config, similar to what mbsync does?

tadfisher: It's probably best you leave this as-is, then. I was just considering the home-manager service module, which needs to set up a systemd timer per-account. However, it's best that way, since you could configure options and timer intervals per-account.

lieer/local.py
NickHu: I've put in a try-catch block to address this.

gauteh:
Does running `gmi init` automatically add a section to the global configuration file (then perhaps the first steps could be removed and we would end up with something similar to previous behavior, I think this would be simper for the user)? What is the difference between `gmi init` and `gmi auth` in the repository now? Can we make `gmi sync` etc without an account argument try to use the current pwd and look it up in the config. This should make this partially backwards compatible. Perhaps the config could be migrated the first time this is run if there is no section in the global config file too.

gauteh:
I think we should either use notmuch config to get maildir_base or use absolute paths for dir. What is the point of this option if it is only used to construct dir? Think we could get away with only one option here.

tadfisher:
@NickHu How's it going?

NickHu:
@tadfisher sorry, it's been busy recently - I do intend to get back to this, hopefully some point in the near future. Alternatively I'm happy for someone else to take over, I believe the rest ought to be fairly simple changes

gauteh:
Closing because of inactivity.

NickHu:
Can't see a way to reopen this so I'll just remake the pull request.

PR #106, opened 2019-04-24
Set mtime of Maildir message to Gmail internalDate

artizirk:
This allows for sorted ls command output and
isync/mbsync use st_mtime when setting IMAP INTERNALDATE

### DIFF
--- a/lieer/local.py
+++ b/lieer/local.py
@@ -421,6 +421,10 @@ class Local:
       with open (tmp_p, 'wb') as fd:
         fd.write (msg_str)
 
+      # Set atime and mtime of the message file to Gmail receive date
+      internalDate = int(m['internalDate']) / 1000  # ms to s
+      os.utime(tmp_p, (internalDate, internalDate))
+
       os.rename (tmp_p, p)
 
     # add to notmuch


gauteh:
Nice, this looks pretty straight forward. Will try to find time to test as soon as possible.

gauteh:
Merged.

dtzWill:
I haven't seen this cause any problems, but just wondering since it has been nagging me: `notmuch-new` makes use of mtime's of files and directories, although I'm not sure of the details.  This change won't interfere with those checks, right? Just checking we're not possibly back-dating mail in a way that it never is found, or something...

Seems to work for me, and it's a nice improvement! Thanks!

gauteh:
Good! Gmaillier adds the messages to the notmuch dB, so running notmuch new
should not have any effect (nor be necessary).

fre. 10. mai 2019 kl. 13:48 skrev Will Dietz <notifications@github.com>:

dtzWill:
On Fri, 10 May 2019 07:08:44 -0700, Gaute Hope <notifications@github.com> wrote:
> Good! Gmaillier adds the messages to the notmuch dB, so running notmuch new
> should not have any effect (nor be necessary).

Great! Thanks for confirming, on both counts actually :).
I wasn't running `notmuch new` but good to know that's right.

FWIW the docs (docs/index.md) mention running `notmuch new` to trigger
the post-new hook, which suggests it may have an effect?
(but not the primary effect it usually does: discovering+adding mail).
No problem if that's accurate but given your statement above
thought I'd double-check ;).

(Sorry for hijacking this PR!)

gauteh:
That’s true, gmailieer adds the new tags so that post new hooks will have
those available 😊

lør. 11. mai 2019 kl. 01:57 skrev Will Dietz <notifications@github.com>:

PR #115, opened 2019-08-29
Fix expand ignore doc

### LINKED ISSUE #94: suggest notmuch ignore pattern is invalid
Concretely, using the provided `ignore=*.json;` didn't result in notmuch skipping json files, which it would complain about when executing `notmuch new`.

While investigating, it seems:

* Trailing semicolon should be removed-- as-is it encodes the list of patterns: `[ "*.json" "" ]` with the second pattern being empty string which is invalid.
* No mention of support for globbing as the pattern suggests.  0.26 introduced support for regex, but the provided pattern is neither delimited with slashes as required nor is it a valid regex (closest would be `.*\.json` or so).

While I quite possibly am doing something wrong since this seems to work for everyone else (?),
the following seems to work but may be incomplete or imprecise-- I'm still trying to get my initial sync to complete so I mostly wanted something good enough to be sure notmuch wasn't interfering with files it shouldn't.

Pattern that works but probably should be reviewed: `/.*.json$/` (usage would be `ignore=/.*.json$/`).

Actually looking at it now I wasn't able to sort out how to escape a period so just punted on that now--originally attempting things like `/\..*\.json$/`.

----

On the subject, ignoring json files seems insufficient? In my limited usage so far I've found gmailieer creates other files resulting in ignoring the following as well:

`/.*.json$/;.gmailieer.json;.gmailieer.json.bak;.lock`

Apologies if I'm missing something, otherwise hope this helps improve things for folks! Thanks!

nickurak:
Correct documented ignore regex and adds additional suffixes.

Fixes #94

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -48,7 +48,7 @@ All commands should be run from the local mail repository unless otherwise speci
 ```
 [new]
 tags=new
-ignore=*.json;
+ignore=/.*[.]json$/
 ```
 
 3. Initialize the mail storage:
--- a/docs/index.md
+++ b/docs/index.md
@@ -48,7 +48,7 @@ All commands should be run from the local mail repository unless otherwise speci
 ```
 [new]
 tags=new
-ignore=/.*[.]json$/
+ignore=/.*[.](json|lock|bak)$/
 ```
 
 3. Initialize the mail storage:


gauteh:
Looks good, thanks!

PR #117, opened 2019-09-05
Separate configuration and state

NickHu:
Rebased onto master. I have time to work on this now, what more needs to be done to #105 for you to merge?

I have been using it for months now with no issues.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -30,20 +30,29 @@ After cloning this repository, symlink `gmi` to somewhere on your path, or use `
 
 # Usage
 
-This assumes your root mail folder is in `~/.mail` and that this folder is _already_ set up with notmuch.
+This assumes your root mail folder is _already_ set up with notmuch.
 
-1. Make a directory for the lieer storage and state files (local repository).
+1. Configure the base mail directory (defaults to `~/.mail`), in
+   `$XDG_CONFIG_HOME/lieer/config`:
 
-```sh
-$ cd    ~/.mail
-$ mkdir account.gmail
-$ cd    account.gmail/
+```ini
+[DEFAULT]
+maildir_base = ~/Mail
 ```
 
-All commands should be run from the local mail repository unless otherwise specified.
+2. Optionally set up accounts to be stored in specific directories under
+   `maildir_base` (defaults to account name, i.e.\ email address), in
+   `$XDG_CONFIG_HOME/gmailieer/config`:
 
+```ini
+[your.email@gmail.com]
+dir = personal_mail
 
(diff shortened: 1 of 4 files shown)


gauteh:
Hi, thanks for picking this up again. Would it work for the original purpose of this PR to have the (immutable) configuration files also in the same directory as the state? So as to avoid the account argument, but rather use context. Maybe use a `gmi -C path/to/repo` and/or `gmi -c path/to/config`? Defaulting to current dir?

One main reason for doing tha is that that is pretty similar to how it currently works. And secondly, it feels less constrained to have multiple totally independent files for each context (similar to git). I realise this is a matter of taste, mostly.

I could not immediately see migration code, how would that work? And we should probably write some tests for that. I think many users use this in an automated setup and would not necessarily pick it up if gmi fails in the background. So we have to be pretty certain that the migration works. This change is pretty risky from that point of view.

NickHu:
No, I think that defeats the purpose of this pull request - the whole idea is to have a configuration file that is in a predictable place which can be statically configured with something like nix. That shouldn't be predicated on however someone has notmuch setup.

I'm in favour of the current implementation. If you feel that migration code is necessary then perhaps we could look at how that could be done but the cleanest way is probably to throw in a deprecation notice in there too so that we can simplify the codebase at some point in the future.

gauteh:
Is nix and home-manager the same thing in this context? Breaking without migration is a very big step. By the way, no matter how careful we do this it will surely generate several new issues. Is this something you are prepared to help out with if we reach a consensus?

With the directory-context way (current) we can make an automatic migration by splitting the files, this seems like something we could be able to do with little migration code and relatively safely. With the proposed changes migration setup will be more complicated (guessing account from current directory, what if gmi is run with account arg in a different account dir).

I think my biggest concern is that changing this behavior now is a big step. If we were designing from scratch your setup would have been at least as good.

tadfisher:
`nix` is just a build tool; given a Nix script which defines inputs and a build recipe, it builds outputs and stores them in the Nix store. This isn't very useful for system configuration, so NixOS and home-manager are tools that take these outputs and, in the typical case, symlink them into the appropriate locations to create a usable configuration. `nix` doesn't know or care about where it outputs end up, but home-manager/NixOS need to know where to place them. In any case, this point is moot, because lieer's configuration lives in a well-known place for a given account, and I can reuse the existing notmuch configuration to determine where lieer's config resides.

To be honest, I feel like this solution is over-engineered. Here's a description of a solution that would only minimally impact existing users, while still allowing for a static configuration file:

- lieer continues to read static configuration from `$maildir/.gmailieer.json`.
- lieer begins writing mutable state to, e.g., `$maildir/.gmailieer.state.json`.
- When reading state, lieer first attempts to read from `$maildir/.gmailieer.state.json`, falling back to `$maildir/.gmailieer.json` if it doesn't exist.

With this behavior, existing users aren't impacted, configuration is silently migrated with no user interaction, and there are no changes to how users target accounts with the `gmi` command.

gauteh:
Thanks. #118 has been merged, closing this one.

PR #118, opened 2019-09-13
Pull static config into Config object

tadfisher:
This is a simpler implementation of #117.

- Pull all static config values into a `Local.Config` class. `Local.State` now just contains `last_historyId` and `lastmod`.
- `Local.Config` loads from `.gmailieer.json` as before.
- `Local.State` first attempts to load from `.state.gmailieer.json`, falling back to `.gmailieer.json` if the state file doesn't exist.
- `Local.State` always writes to `.state.gmailieer.json`.
- `Local` now has a `config` field, and everything referencing the old values in `state` is now referencing those same values from `config`.

I am currently using this and it appears to work fine. The old `last_historyId` and `lastmod` values stick around in `.gmailieer.json` until the config is modified.

### DIFF
--- a/lieer/local.py
+++ b/lieer/local.py
@@ -51,16 +51,8 @@ class Local:
   class RepositoryException (Exception):
     pass
 
-  class State:
-    # last historyid of last synchronized message, anything that has happened
-    # remotely after this needs to be synchronized. gmail may return a 404 error
-    # if the history records have been deleted, in which case we have to do a full
-    # sync.
-    last_historyId = 0
 
-    # this is the last modification id of the notmuch db when the previous push was completed.
-    lastmod = 0
-
+  class Config:
     replace_slash_with_dot = False
     account = None
     timeout = 0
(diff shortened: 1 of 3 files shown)


lieer/local.py
```diff
@@ -159,13 +139,62 @@ def set_file_extension (self, t):
...
+
+    def __init__ (self, state_f, config_f):
+      self.state_f = state_f
+
+      if os.path.exists (self.state_f):
+        with open (self.state_f, 'r') as fd:
+          self.json = json.load (fd)
+      elif os.path.exists (config_f):
+        with open (config_f, 'r') as fd:
+          self.json = json.load (fd)
```
gauteh: One thing I thought of here is if the user tries to delete the state file just to do a clean sync, then we will pull in the settings from the config file. Perhaps we should clear those values from the config file if they were loaded from there (on the next write to the state).. maybe I am overcomplicating things now.

tadfisher: Probably a good idea. We could write the config file immediately if it contains one of the state keys.

gauteh:
Thanks. I do feel more comfortable with this approach over #117, though I appreciate the work done there and think that it could very well have been a better solution if we were starting out earlier in the projects history. Any comments @NickHu?

NickHu:
I'm happy to go with @tadfisher's approach - he wrote the home-manager module so if it's good enough for him, it's good enough for me

lieer/local.py
gauteh: Sounds good. I think with that in place this should be pretty much good to go.

tadfisher: The latest commit takes a slightly more conservative approach, where during state load, both `config` and `state` are written immediately if the following conditions apply:

- A state file doesn't exist
- A config file does exist
- The config file contains `last_historyId` or `lastmod` keys

gauteh: Looks good!

gauteh:
Thanks, merged :)

NickHu:
Now that this is merged, and a new release of lieer has been issued, is it perhaps time to upstream https://github.com/tadfisher/home-manager/tree/gmailieer?

PR #123, opened 2019-11-27
Give a hint where to find notmuch config to ignore json files

mrowe:
I was confused for a while about where I should be making the change to tell notmuch to ignore gmi json files (and stop applying the inbox tag to everything!). Perhaps others will be too, so make it a bit more explicit.


gauteh:
Looks good, merged!

PR #124, opened 2019-11-28
Explicit error message if json files fails to load.

aureooms:
Currently, the error message does not explicitly mention the file that failed to load (incorrect JSON encoding).


gauteh:
Thanks, merged!

PR #129, opened 2020-03-12
Allow to only fetch recent emails

### LINKED ISSUE #93: only fetch recent mail (limit disk size)
Some IMAP fetching solutions offer the capability to limit what mail is fetched based on criteria like message (or thread) age, or by an on-disk size limit.

Is there a way to do this with gmailieer? Maybe by playing games with historyId?

My use case is that I'm low on disk space on my laptop and I would love to only have the previous 30 days of mail sync'd. On my workstation I want everything sync'd (the current default behavior of gmailieer).

Any ideas? Is this already supported?

coiby:
This is an partial implementation of [the `newer_than` search operator](https://support.google.com/mail/answer/7190?hl=en). Only emails newer than specified months will be fetched. This feature will fix #93.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -111,7 +111,7 @@ Lieer can be configured using `gmi set`. Use without any options to get a list o
 
 **`historyId`** is the latest synced GMail revision. Anything since this ID will be fetched on the next [`gmi pull`](#pull) (partial).
 
-**`age`** is the maximum age of email. This is an partial implementation of [the `newer_than` search operator](https://support.google.com/mail/answer/7190?hl=en). Any email newer than this age (unit: month) will be fetched on the next [`gmi pull`](#pull). If you change this parameter to an older age, you need to do a full pull.
+**`age`** is the maximum age of email (unit: month). This is an partial implementation of [the `newer_than` search operator](https://support.google.com/mail/answer/7190?hl=en). Any email newer than this age will be fetched on the next [`gmi pull`](#pull). If you change this parameter to an older age, you need to do a full pull. *Note that this is an experimental feature for now*. If necessary, you can use `gmi set --unset-age` to remove it from the configuration.
 
 **`lastmod`** is the latest synced Notmuch database revision. Anything changed after this revision will be pushed on [`gmi push`](#ush).
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -111,11 +111,13 @@ Lieer can be configured using `gmi set`. Use without any options to get a list o
 
 **`historyId`** is the latest synced GMail revision. Anything since this ID will be fetched on the next [`gmi pull`](#pull) (partial).
 
+**`age`** is the maximum age of email. This is an partial implementation of [the `newer_than` search operator](https://support.google.com/mail/answer/7190?hl=en). Anything email newer than this age (unit: month) will be fetched on the next [`gmi pull`](#pull). If you change this parameter to an older age, you need to do a full pull.
+
 **`lastmod`** is the latest synced Notmuch database revision. Anything changed after this revision will be pushed on [`gmi push`](#ush).
 
 **`Timeout`** is the timeout in seconds used for the HTTP connection to GMail. `0` means the forever or system error/timeout, [whichever occurs first](https://github.com/gauteh/lieer/issues/83#issuecomment-396487919).
 
-**`File extension`** is an optional argument to include the specified extension in local file names (e.g., `mbox`) which can be useful for indexing them with third-party programs.  
+**`File extension`** is an optional argument to include the specified extension in local file names (e.g., `mbox`) which can be useful for indexing them with third-party programs.
 
 *Important:* If you change this setting after synchronizing, the best case scenario is that all files will apear to not have being pulled down and will be re-downloaded (and duplicated with a different extension in the maildir). There might also be changes to tags. You should in theory be able to change it by renaming all files, but since this will update the lastmod you will get a check on all files.
 
@@ -124,11 +126,11 @@ Lieer can be configured using `gmi set`. Use without any options to get a list o
 **`Replace slash with dot`** is used to replace the sub-label separator (`/`) with a dot (`.`). I think this is easier to work with. *Important*: See note below on [changing this setting after initial sync](#changing-ignored-tags-and-translation-after-initial-sync).
 
 **`Ignore tags (local)`** can be used to specify a list of tags which should not be synced from local to remote (e.g. [`new`](#usage)). In addition to the user-configured tags these tags are ignored: `'attachment', 'encrypted', 'signed', 'passed', 'replied', 'muted', 'mute', 'todo', 'Trash', 'voicemail'`. Some are special tags in notmuch and some are unsupported by GMail. See [Caveats](#caveats) below for more explanations. *Note:* This setting expects [_translated_ tags](#translation-between-labels-and-tags).
-  
+
   *Important*: See note below on [changing this setting after initial sync](#changing-ignored-tags-and-translation-after-initial-sync).
 
 **`Ignore tags (remote)`** can be used to specify a list of tags (labels) which should not be synced from remote (GMail) to local. By default the [`CATEGORY_*` type](https://developers.google.com/gmail/api/guides/labels) labels which are mapped to the Personal/Promotions/etc tabs in the GMail interface are ignored. You can specify that no label should ignored by doing: `gmi set --ignore-tags-remote ""`. *Note:* This setting expects [_*un*translated_ tags](#translation-between-labels-and-tags).
-  
+
   *Important*: See note below on [changing this setting after initial sync](#changing-ignored-tags-and-translation-after-initial-sync).
 
 
--- a/docs/index.md
+++ b/docs/index.md
@@ -111,7 +111,7 @@ Lieer can be configured using `gmi set`. Use without any options to get a list o
 
 **`historyId`** is the latest synced GMail revision. Anything since this ID will be fetched on the next [`gmi pull`](#pull) (partial).
 
-**`age`** is the maximum age of email. This is an partial implementation of [the `newer_than` search operator](https://support.google.com/mail/answer/7190?hl=en). Anything email newer than this age (unit: month) will be fetched on the next [`gmi pull`](#pull). If you change this parameter to an older age, you need to do a full pull.
+**`age`** is the maximum age of email. This is an partial implementation of [the `newer_than` search operator](https://support.google.com/mail/answer/7190?hl=en). Any email newer than this age (unit: month) will be fetched on the next [`gmi pull`](#pull). If you change this parameter to an older age, you need to do a full pull.
 
 **`lastmod`** is the latest synced Notmuch database revision. Anything changed after this revision will be pushed on [`gmi push`](#ush).
 
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -45,9 +45,6 @@ class Gmailieer:
     parser_pull.add_argument ('--limit', type = int, default = None,
         help = 'Maximum number of messages to pull (soft limit, GMail may return more), note that this may upset the tally of synchronized messages.')
 
-    parser_pull.add_argument ('--age', type = int, default = None,
-        help = 'Maximum age (unit: month) of messages to pull')
-
     parser_pull.add_argument ('-d', '--dry-run', action='store_true',
         default = False, help = 'do not make any changes')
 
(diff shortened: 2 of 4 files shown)


lieer/gmailieer.py
```diff
@@ -191,6 +202,7 @@ def setup (self, args, dry_run = False, load = False):
     self.dry_run          = dry_run
     self.HAS_TQDM         = (not args.no_progress)
     self.credentials_file = args.credentials
+    self.age = args.age if hasattr(args, 'age') else None
```
gauteh: Can this be just `args.age`?

docs/index.md
```diff
@@ -124,11 +126,11 @@ Lieer can be configured using `gmi set`. Use without any options to get a list o
 **`Replace slash with dot`** is used to replace the sub-label separator (`/`) with a dot (`.`). I think this is easier to work with. *Important*: See note below on [changing this setting after initial sync](#changing-ignored-tags-and-translation-after-initial-sync).
 
 **`Ignore tags (local)`** can be used to specify a list of tags which should not be synced from local to remote (e.g. [`new`](#usage)). In addition to the user-configured tags these tags are ignored: `'attachment', 'encrypted', 'signed', 'passed', 'replied', 'muted', 'mute', 'todo', 'Trash', 'voicemail'`. Some are special tags in notmuch and some are unsupported by GMail. See [Caveats](#caveats) below for more explanations. *Note:* This setting expects [_translated_ tags](#translation-between-labels-and-tags).
-  
+
```
gauteh: These whitespaces are there to create new paragraphs and should be left in.

lieer/remote.py
```diff
@@ -106,6 +106,8 @@ def __init__ (self, g):
     self.CLIENT_SECRET_FILE = g.credentials_file
     self.account = g.local.config.account
     self.dry_run = g.dry_run
+    if g.local.age:
+       self.query = "newer_than:{}m".format(g.local.age)
```
gauteh: This resets the existing query, it should be appended with an AND.

docs/index.md
```diff
@@ -111,11 +111,13 @@ Lieer can be configured using `gmi set`. Use without any options to get a list o
 
 **`historyId`** is the latest synced GMail revision. Anything since this ID will be fetched on the next [`gmi pull`](#pull) (partial).
 
+**`age`** is the maximum age of email. This is an partial implementation of [the `newer_than` search operator](https://support.google.com/mail/answer/7190?hl=en). Anything email newer than this age (unit: month) will be fetched on the next [`gmi pull`](#pull). If you change this parameter to an older age, you need to do a full pull.
+
```
gauteh: `Anything` -> `Any`. Will this cause emails to be deleted (locally) when they go out of the newer_than window? This should probably refer to the section about changing some crucial config settings.

lieer/gmailieer.py
coiby: I'm not sure if it make sense to use age parameter for push action. For now args.age doesn't exist for push action.

docs/index.md
coiby: Sorry, these whitespaces are automatically deleted by vim when saving the file. But don't we only need newline to create a new paragraph?

lieer/remote.py
coiby: Thank you for reminding me of this issue. I'll fix it in the next commit.

docs/index.md
gauteh: No, I think they are needed. But if it can be solved with an extra newline  then you could try that. Should be observable on your branch on github.

lieer/gmailieer.py
gauteh: I think the parameter should only be available on the `set` command. Having the above line in def setup() and in the action as well is redundant, it should only be on the action unless it is a command argument.

docs/index.md
coiby: Thank you for reminding me of this typo. Based on my understanding of the lieer code and the Gmail API docs, setting age to a smaller number will cause emails to be deleted which has been confirmed by a manual test.

lieer/gmailieer.py
coiby: As I review your last feedback, I found only messages.list and threds.list method support this query. According to https://developers.google.com/gmail/api/guides/filtering,
> You can search or filter files using the messages.list and threads.list methods. These methods accept the q parameter which supports the same advanced search syntax as the Gmail web-interface. 

So args.age doesn't make sense for push action.

coiby:
On Sat, Mar 14, 2020 at 04:13:12AM -0700, Gaute Hope wrote:
>> @@ -191,6 +202,7 @@ def setup (self, args, dry_run = False, load = False):
>     self.dry_run          = dry_run
>     self.HAS_TQDM         = (not args.no_progress)
>     self.credentials_file = args.credentials
>+    self.age = args.age if hasattr(args, 'age') else None
>
>I think the parameter should only be available on the `set` command. Having the above line in def setup() and in the action as well is redundant, it should only be on the action unless it is a command argument.

  Are you suggesting dropping support for `gmi pull/init --age N` and only supporting `gmi set --age N`? I'm bit confused about the terminology and not sure if `pull` should be called an action.

coiby:
To be precise, two newlines start a new paragraph according to the markdown syntax. My branch is identical to your branch except for the new paragraph about `age`. So extra spaces are not necessary.

gauteh:
On Sat, 14 Mar 2020 at 13:54 Coiby wrote:
> To be precise, two newlines start a new paragraph according to the markdown syntax. My branch is identical to your branch except for the new paragraph about `age`. So extra spaces are not necessary.
>

That sounds good.

Is it possible to unset the age parameter (back to None)?

Will local files be deleted unless you do a full pull? The purpose of
the feature is to avoid having all email locally? If it only works on
full pull it will most often not actually take effect?

I think this option has potential for messing up tags and content quite
a bit. It should be tested a fair bit before merging. I think it should
be marked as experimental in the beginning.

-- gaute

coiby:
>Is it possible to unset the age parameter (back to None)?

I'll add some code to set the age parameter to be None if `gmi set --age
n` where n is an integer is a negative number in the next commit.


>Will local files be deleted unless you do a full pull? The purpose of
>the feature is to avoid having all email locally? If it only works on
>full pull it will most often not actually take effect?

Yes, the purpose is to fetch only parts of emails because a full fetch is
time-consuming for me and also disk-consuming for other people.

Both full and partial pull will not delete local files.

For full pull, unless
`--remove` option is specified, local emails will not be deleted,
```python
   def full_pull (self):
     ...
     if self.remove:
       if self.limit and not self.dry_run:
         raise argparse.ArgumentError ('--limit with --remove will cause lots of messages to be deleted')

       # removing files that have been deleted remotely
       all_remote = set (message_gids)
       all_local  = set (self.local.gids.keys ())
       remove     = list(all_local - all_remote)
       bar = tqdm (leave = True, total = len(remove), desc = 'removing deleted')
       with notmuch.Database (mode = notmuch.Database.MODE.READ_WRITE) as db:
         for m in remove:
           self.local.remove (m, db)
```

And for `partial_pull`, old emails will still be preserved locally. If we
change the age parameter to an later date, it will not delete old local
files since only newer emails have the `messagesDeleted` label,
```python
   def partial_pull(self):
     ...
     for h in history:
       if 'messagesDeleted' in h:
         for m in h['messagesDeleted']:
           mm = m['message']
           # might silently fail to delete this
           remove_from_all (mm)
           if self.local.has (mm['id']):
             deleted_messages.append (mm)
```

>I think this option has potential for messing up tags and content quite
>a bit. It should be tested a fair bit before merging. I think it should
>be marked as experimental in the beginning.

Although I've tested it locally by changing age to a late date, it's
safe than sorry. So it's reasonable to mark it as experimental in the
beginning.

gauteh:
Coiby writes on March 24, 2020 12:24:
>>Is it possible to unset the age parameter (back to None)?
>
> I'll add some code to set the age parameter to be None if `gmi set --age
> n` where n is an integer is a negative number in the next commit.

Great. Best to stay with current convention of adding an --unset-age
type option like the existing options.

>>Will local files be deleted unless you do a full pull? The purpose of
>>the feature is to avoid having all email locally? If it only works on
>>full pull it will most often not actually take effect?
>
> Yes, the purpose is to fetch only parts of emails because a full fetch is
> time-consuming for me and also disk-consuming for other people.
>
> Both full and partial pull will not delete local files.
>
> For full pull, unless
> `--remove` option is specified, local emails will not be deleted,

Ok. I think this still makes sense.

> And for `partial_pull`, old emails will still be preserved locally. If we
> change the age parameter to an later date, it will not delete old local
> files since only newer emails have the `messagesDeleted` label,
> ```python
> def partial_pull(self):
> ...
> for h in history:
> if 'messagesDeleted' in h:
> for m in h['messagesDeleted']:
> mm = m['message']
> # might silently fail to delete this
> remove_from_all (mm)
> if self.local.has (mm['id']):
> deleted_messages.append (mm)
> ```

Good. That looks good. We should add a comment about this to the
'messageDeleted' code since it is not so obvious, in case we change
that code in the future.

>>I think this option has potential for messing up tags and content quite
>>a bit. It should be tested a fair bit before merging. I think it should
>>be marked as experimental in the beginning.
>
> Although I've tested it locally by changing age to a late date, it's
> safe than sorry. So it's reasonable to mark it as experimental in the
> beginning.

coiby:
On Tue, Mar 24, 2020 at 04:31:35AM -0700, Gaute Hope wrote:
>Coiby writes on March 24, 2020 12:24:
>>>Is it possible to unset the age parameter (back to None)?
>>
>> I'll add some code to set the age parameter to be None if `gmi set --age
>> n` where n is an integer is a negative number in the next commit.
>
>Great. Best to stay with current convention of adding an --unset-age
>type option like the existing options.

Could you give me an example which has an --unset-age type option? I
couldn't find out one and the docs of ArgumentParser.add_argument()
don't seem to provide related description.

gauteh:
Coiby writes on March 24, 2020 13:06:
> On Tue, Mar 24, 2020 at 04:31:35AM -0700, Gaute Hope wrote:
>>Coiby writes on March 24, 2020 12:24:
>>>>Is it possible to unset the age parameter (back to None)?
>>>
>>> I'll add some code to set the age parameter to be None if `gmi set --age
>>> n` where n is an integer is a negative number in the next commit.
>>
>>Great. Best to stay with current convention of adding an --unset-age
>>type option like the existing options.
>
> Could you give me an example which has an --unset-age type option? I
> couldn't find out one and the docs of ArgumentParser.add_argument()
> don't seem to provide related description.

https://github.com/gauteh/lieer/blob/master/lieer/gmailieer.py#L145

coiby:
On Wed, Mar 25, 2020 at 02:53:11PM +0000, Gaute Hope wrote:
>Coiby writes on March 24, 2020 13:06:
>> On Tue, Mar 24, 2020 at 04:31:35AM -0700, Gaute Hope wrote:
>>>Coiby writes on March 24, 2020 12:24:
>>>>>Is it possible to unset the age parameter (back to None)?
>>>>
>>>> I'll add some code to set the age parameter to be None if `gmi set --age
>>>> n` where n is an integer is a negative number in the next commit.
>>>
>>>Great. Best to stay with current convention of adding an --unset-age
>>>type option like the existing options.
>>
>> Could you give me an example which has an --unset-age type option? I
>> couldn't find out one and the docs of ArgumentParser.add_argument()
>> don't seem to provide related description.
>
>https://github.com/gauteh/lieer/blob/master/lieer/gmailieer.py#L145

Thank you! Done in the latest commit.

gauteh:
Hi @coiby, I have thought a bit about this feature. And I do not feel confident adding it as it is. It adds a fair bit of complexity, and I feel that there are possible side-effects and scenarios that could potentially cause data-loss or inconsistencies that we (or at least I)  do not have the full overview of. Relative to the number of users that would use this feature, I therefore think that this should not be included.

frobware:
> Hi @coiby, I have thought a bit about this feature. And I do not feel confident adding it as it is. It adds a fair bit of complexity, and I feel that there are possible side-effects and scenarios that could potentially cause data-loss or inconsistencies that we (or at least I) do not have the full overview of. Relative to the number of users that would use this feature, I therefore think that this should not be included.

Is there an alternative means of syncing less? Every time I've played with this utility I run into quotas and have quite often locked myself out of my gmail account for 24 hours. I wanted to try again today -- noticed this feature -- and thought "great, maybe syncing less stuff will allow me to experiment some more."

gauteh:
Andrew McDermott writes on May 15, 2020 0:21:
> Is there an alternative means of syncing less? Every time I've played
> with this utility I run into quotas and have quite often locked myself
> out of my gmail account for 24 hours. I wanted to try again today --
> noticed this feature -- and thought "great, maybe syncing less stuff
> will allow me to experiment some more."

You could give this PR a shot. I never run into quota issues once I have
synced the emails the first time. Usually never the first time either.
There is a ratio of how many requests per second you can use, but lieer
should back-off if that occurs.

PR #131, opened 2020-03-19
send-message

gauteh:


julian-klode:
Message sending is a bit nasty in my experience, as it does not link messages you send into conversations itself, you have to explicitly set the conversation id or what its called so that stuff appears correctly in gmail.

gauteh:
It is possible to pass that with the API, but would require it to be passed by the sender. Does it make a difference from using an SMTP client?

julian-klode:
Yeah, if you send via smtp, gmail figures out the conversation it belongs to automatically.

gauteh:
Now uses in-reply-to to try and figure out threadId.

gauteh:
Quoting Julian Andres Klode (2020-03-19 16:52:13)
> Yeah, if you send via smtp, gmail figures out the conversation it belongs to automatically.

Seems to work alright, though not extensively tested yet.

gauteh:
Here's a sendmail line:

```
gmi send -C ~/.mail/gmail.account
```

PR #143, opened 2020-05-05
Expand ~user for -C argument (fix #141)

### LINKED ISSUE #141: Make -C expand ~user
When passing a path to lieer using -C from a programming language, rather than a shell, it would be nice to be able to use "~" in the path and have it expanded. I get:

```
path: ~/.mail/account.gmail
error: ~/.mail/account.gmail is not a valid path!
  File "/home/skangas/src/lieer/gmi", line 8, in <module>
    g.main ()
  File "/home/skangas/src/lieer/lieer/gmailieer.py", line 182, in main
    args.func (args)
  File "/home/skangas/src/lieer/lieer/gmailieer.py", line 681, in send
    self.setup (args, args.dry_run, True, True)
  File "/home/skangas/src/lieer/lieer/gmailieer.py", line 223, in setup
    raise NotADirectoryError("error: %s is not a valid path!" % args.path)
NotADirectoryError: error: ~/.mail/account.gmail is not a valid path!
```

This is just a small quality of life improvement.  Thanks.

skangas:
How about this?

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -205,6 +205,7 @@ class Gmailieer:
       if args.action == "init" and not os.path.exists(args.path):
         os.makedirs(args.path)
 
+      args.path = os.path.expanduser(args.path)
       if os.path.isdir(args.path):
         self.cwd = os.getcwd()
         os.chdir(args.path)


keegancsmith:
@skangas why can't you rely on your shell to do this expansion? Normally you use expanduser when reading variables that don't come from your shell (eg a configuration file)

gauteh:
Is this required for e.g. emacs to work with ~? We should not add
functionality that is not necessary, but on the other hand I do not see
any big problems with including this.

keegancsmith:
> Is this required for e.g. emacs to work with ~?

In emacs you can just use (expand-file-name "~/my/path").

> We should not add functionality that is not necessary, but on the
> other hand I do not see any big problems with including this.

I agree. Not a problem but always nice to avoid unnecessary
code/behaviour.

skangas:
As I explained in #143 it's a quality of life improvement. lieer is not always called through the shell, so ~user is not always expanded.

In my use-case, GNU Emacs, the difference is having to type:  

```
(setq message-sendmail-extra-arguments `("send" "-C" (expand-file-name "~/.mail/account.gmail")))
```

and being able to say

```
(setq message-sendmail-extra-arguments `("send" "-C" "~/.mail/account.gmail"))
```

I think supporting the latter makes for less surprises. As we know, ~user is strictly for convenience and not having to modify your configuration for every particular machine.

gauteh:
Merged :)

skangas:
Thanks.

PR #144, opened 2020-05-06
Add --quiet option for sendmail compatibility

skangas:
There is one final complication to use lieer as a sendmail stand-in from GNU Emacs. Emacs treats _any_ output from sendmail as an error. I personally think lieer could say very little by default, which is in line with the Unix tradition and how sendmail works traditionally, but I'd like to suggest here a less intrusive alternative:

Default output stays as it is, but we add a new --quiet option.

(I have only tweaked this in detail for the sync and send actions.)

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -28,6 +28,9 @@ class Gmailieer:
     common.add_argument ('-s', '--no-progress', action = 'store_true',
         default = False, help = 'Disable progressbar (always off when output is not TTY)')
 
+    common.add_argument ('-q', '--quiet', action = 'store_true',
+        default = False, help = 'Produce less output (implies -s)')
+
     subparsers = parser.add_subparsers (help = 'actions', dest = 'action')
     subparsers.required = True
 
@@ -181,6 +184,9 @@ class Gmailieer:
     args        = parser.parse_args (sys.argv[1:])
     self.args   = args
 
+    if args.quiet:
+      args.no_progress = True
+
     args.func (args)
 
   def initialize (self, args):
@@ -213,7 +219,7 @@ class Gmailieer:
 
     # common options
     if args.path is not None:
-      print("path: %s" % args.path)
+      self.vprint ("path: %s" % args.path)
       if args.action == "init" and not os.path.exists(args.path):
         os.makedirs(args.path)
 
@@ -280,7 +286,7 @@ class Gmailieer:
       (rev, uuid) = db.get_revision ()
 
       if rev == self.local.state.lastmod:
-        print ("push: everything is up-to-date.")
+        self.vprint ("push: everything is up-to-date.")
         return
 
       qry = "path:%s/** and lastmod:%d..%d" % (self.local.nm_relative, self.local.state.lastmod, rev)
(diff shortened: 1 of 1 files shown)


lieer/gmailieer.py
```diff
@@ -699,33 +708,33 @@ def send (self, args):
...
       repl = eml['In-Reply-To'].strip().strip('<>')
-      print("looking for original message: %s" % repl)
+      self.vprint("looking for original message: %s" % repl)
       with notmuch.Database (mode = notmuch.Database.MODE.READ_ONLY) as db:
         nmsg = db.find_message(repl)
         if nmsg is not None:
           (_, gids) = self.local.messages_to_gids([nmsg])
           if nmsg.get_header('Subject') != eml['Subject']:
-            print("warning: subject does not match, might not be able to associate with existing thread.")
+            self.vprint ("warning: subject does not match, might not be able to associate with existing thread.")
```
gauteh, line 721: Maybe this should be `print` too? It is a warning, but less serious than the those that are still `print` below.

lieer/gmailieer.py
```diff
@@ -775,3 +784,30 @@ def set (self, args):
 
 
 
+  def vprint (self, *args):
+    """
+    Print unless --quiet.
+    """
+    if not self.args.quiet:
+      print (*args)
```
gauteh, line 792: maybe include  `kwargs`

gauteh:
Looks good, I think there might be some statements in remote and local that are not strictly necessary either. Some of those are related to initialization, which should maybe be kept. And some are warnings/errors that should be included.

lieer/gmailieer.py
skangas: I find that it's emitted often (or all the time, I'm not sure) when sending, but I'm not sure why exactly. This makes Emacs complain that sending the message failed.

So to make this work, we need to either silence it or figure out why it's getting printed and stop that situation from occurring. Either on Emacs' side or lieer's.

lieer/gmailieer.py
skangas: Sure, I'll add that.

skangas:
> Looks good, I think there might be some statements in remote and local that are not strictly necessary either. Some of those are related to initialization, which should maybe be kept. And some are warnings/errors that should be included.

My preference, especially for --quiet, is to only emit serious warnings or actual errors. I can have a look at those actions and see if I can come up with an acceptable set of suggestions.

lieer/gmailieer.py
gauteh: The google API docs require the subject to match, but in reality it does not seem to be required (at least not if it is just prepending a Re:). I have not fixed it in e.g. astroid so that the subject remains identical. I have documented it in the README, so it should be fine to quiet this warning.

gauteh:
> > Looks good, I think there might be some statements in remote and local that are not strictly necessary either. Some of those are related to initialization, which should maybe be kept. And some are warnings/errors that should be included.
> 
> My preference, especially for --quiet, is to only emit serious warnings or actual errors. I can have a look at those actions and see if I can come up with an acceptable set of suggestions.

I agree, I was thinking that maybe more of those should also be quieted (e.g. the exponential backoff).

lieer/gmailieer.py
skangas: Thank you for clarifying that.

gauteh: Just realized that I had not documented it, but since it does not seem to be a real requirement we could maybe even just drop the warning. Or drop the warning and document it.

skangas:
> I agree, I was thinking that maybe more of those should also be quieted (e.g. the exponential backoff).

Thanks, I see.  Pushing a new patch in a minute, please let me know if you have any further comments.

lieer/gmailieer.py
skangas: Sorry, I missed this comment before pushing. I'm not sure I feel confident making a decision about this or understand the implications.

Maybe it's better to work on that separately? I'm thinking it's better that you write the documentation, than me guessing what it should say...

skangas: In other words, my suggestion would be to merge this and then do whatever improvements are necessary with regards to dropping this warning.

lieer/gmailieer.py
```diff
@@ -299,24 +305,25 @@ def push (self, args, setup = False):
...
+      self.bar_close ()
 
       # resolve changes
-      bar = tqdm (leave = True, total = len(gids), desc = 'resolving changes')
+      self.bar_create (leave = True, total = len(gids), desc = 'resolving changes')
       actions = []
       for rm, nm in zip(remote_messages, messages):
         actions.append (self.remote.update (rm, nm, self.local.state.last_historyId, self.force))
-        bar.update (1)
+        if not self.args.quiet and self.bar:
```
gauteh, line 323: redundant check for `not self.args.quiet`

lieer/gmailieer.py
gauteh: Agree.

lieer/gmailieer.py
skangas: Fixed.

skangas:
Latest version with redundant check removed.

gauteh:
Thanks, merged! Fix to docs also pushed.

skangas:
Thanks for the reviews, and merging!

PR #145, opened 2020-05-07
Document how to use "gmi send" in GNU Emacs

skangas:
Here's the documentation that would help users get started with "gmi send" from GNU Emacs.

Comments or suggestions welcome.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -130,6 +130,8 @@ Note that the following flags are ignored for `sendmail` compatability:
   - `-o` (ignored)
   - `-i` (always implied, not bothered by single `.`'s)
 
+There are instructions for using this in your email client (for example Emacs) on the [wiki](https://github.com/gauteh/lieer/wiki).
+
 # Settings
 
 Lieer can be configured using `gmi set`. Use without any options to get a list of the current settings as well as the current history ID and notmuch revision.


skangas:
Sorry, I accidentally had not rebased on top of master. 00a7e50 is correctly rebased.

gauteh:
Thanks. I was wondering if it might be better to add a link to the README, and put the content in the wiki?

skangas:
> Thanks. I was wondering if it might be better to add a link to the README, and put the content in the wiki?

I think Emacs is an important enough use-case to include in the README, but I might be biased. Please let me know what you prefer.

gauteh:
Stefan Kangas writes on May 7, 2020 12:49:
>> Thanks. I was wondering if it might be better to add a link to the README, and put the content in the wiki?
>
> I think Emacs is an important enough use-case to include in the README, but I might be biased. Please let me know what you prefer.

Hehe, up until now lieer is MTA agnostic, but maybe a link in the
README along the
lines of `See the wiki for how to configure e.g. Emacs...`. ?

skangas:
> Hehe, up until now lieer is MTA agnostic, but maybe a link in the README along the lines of `See the wiki for how to configure e.g. Emacs...`. ?

Thanks, I added a wiki page and referred to it from README.md in the above commit.

gauteh:
Thanks, merged

PR #146, opened 2020-05-07
Add recommended license notice to files

skangas:
[How to use GNU licenses for your own software](https://www.gnu.org/licenses/gpl-howto.html) states:

> Here's a brief summary of what you need to do to release a program under one of our licenses:

>  -    Give each file the proper copyright notices. Make sure to clearly identify which versions of the license users can use.

It's generally good form to follow this advice, in order to make it clear this is free software under the GPL. So I suggest to do it, courtesy of random OCD tendencies. :-)

### DIFF
--- a/gmi
+++ b/gmi
@@ -1,5 +1,21 @@
 #! /usr/bin/env python3
 #
+# Copyright © 2020  Gaute Hope <eg@gaute.vetsj.com>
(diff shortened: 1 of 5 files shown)


gauteh:
Thanks, merged!

PR #147, opened 2020-05-09
Add support for email queue

skangas:
When not using a "real" MTA, it's still a good idea to have a mail queue.  msmtp has msmtpq, for example. So here's an implementation of that for Lieer.

The use-case is that Emacs blocks while handing off the e-mail to `sendmail`, which is not very convenient when `gmi send` can take several seconds to complete (due to Google server latency I suppose).

Please consider merging this, and thanks. Comments are of course very welcome.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -132,6 +132,16 @@ Note that the following flags are ignored for `sendmail` compatability:
 
 There are instructions for using this in your email client (for example Emacs) on the [wiki](https://github.com/gauteh/lieer/wiki).
 
+## Queueing emails
+
+Some mail clients will block while waiting to send an email.  To avoid this, you can instead add it to a queue for batch processing.  You can then simply replace the `gmi send` command with `gmi queue`.  That would typically look like this:
+
+``sh
+gmi queue -C ~/.mail/account.gmail
+``
+
+You can list the contents of the queue using `gmi manage_queue -l', and send the emails in it by running `gmi manage_queue -r`.  See `gmi manage_queue --help` for more options.
+
 # Settings
 
 Lieer can be configured using `gmi set`. Use without any options to get a list of the current settings as well as the current history ID and notmuch revision.
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -26,9 +26,15 @@ import  notmuch
 
 from .remote import *
 from .local  import *
+from .queue  import *
 
 class Gmailieer:
   cwd = None
+
+  QUEUE_DIR = '.lieer.queue'
+
+  def __init__ (self):
+    self.cwd = os.getcwd ()  # save working directory for later
 
   def main (self):
     parser = argparse.ArgumentParser ('gmi', parents = [tools.argparser])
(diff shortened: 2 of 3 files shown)


gauteh:
Wow, nice. I was wondering if this perhaps should be a separate module, I
think it would be nice to have it in the lieer package. But in the spirit
of doing one thing well, I think it should be separated from the current
module. We could provide `gmi queue` as an alias to `gmiq` or whatever.
This would probably be a good point to start using entry-points in
setup.py. In fact `gmi send` should maybe also be separate. Another tool
that I need is contacts through gmail, there is goobook, but I need
something that is account specific and it would be nice to re-use
the existing repo. This should definitely be in a separate module (it needs
a local cache of the contacts).

skangas:
I think that's a good proposal for improving on this work, sure.

I'm a bit pressed for time so I can't promise that I'll be able to get to this any time soon. If anyone else wants to take what I did so far and improve on it, I urge them to please go ahead and not to wait for me. Otherwise, I will address this as time allows.

For what it's worth, I can report that I've been happily using my patch since it was posted with no issues. Thanks.

gauteh:
Stefan Kangas writes on May 12, 2020 18:51:
> I think that's a good proposal for improving on this work, sure.
>
> I'm a bit pressed for time so I can't promise that I'll be able to get to this any time soon. If anyone else wants to take what I did so far and improve on it, I urge them to please go ahead and not to wait for me. Otherwise, I will address this as time allows.
>
> For what it's worth, I can report that I've been happily using my patch since it was posted with no issues. Thanks.

A totally different (and not necessarily better) approach would be to
rely on notmuch-tags for queueing outgoing emails. It would still make
sense to move emails around folders. That would maybe be easier to make
agnostic to which tool you use to send the emails in the end (MTA).

flokli:
Disclaimer: I'm not sure if I'm a fan of this approach in general. 

"Real mailservers" bounce messages back to the sender in case of delivery errors (and as they run on the mailserver, it's somewhat likely at least the bounces are received), or reject a message already before enqueueing it.

While your approach provides some way to inspect the queue contents, it's on the user to notice things are stuck there, and an entirely different interface.

Maintaining the queue inside the notmuch database using tags as suggested by @gauteh would make inspecting the queue easier, but I might personally just stick with the `send` interface, to get instant feedback on possible delivery failures.

skangas:
> Disclaimer: I'm not sure if I'm a fan of this approach in general.
> 
> "Real mailservers" bounce messages back to the sender in case of delivery errors (and as they run on the mailserver, it's somewhat likely at least the bounces are received), or reject a message already before enqueueing it.
> 
> While your approach provides some way to inspect the queue contents, it's on the user to notice things are stuck there, and an entirely different interface.

There is room for improvement, sure. Do you have anything specific in mind?

skangas:
I think the easiest and best solution is to just to print a warning to stdout if there's an error when processing the queue. In my use, I run this from crontab and will get an email if there's any output on stdout. This also suggests that the default mode here should be less verbose.

This is of coulse not how it "should" work, or how email was designed to work. But to be honest, Lieer is already a bit of a hack in that regard as is. Email "should" be handled using real MTA's communicating over standard protocols like IMAP, and not by proprietary Google API:s. But this is because Google is a bit rubbish, and we can't do anything about that. (To be fair, Lieer turns out to be a hack that works well within the given constraints.)

If you want to do more, the easiest step up would likely be to inject local "bounce" messages into the inbox (unfortunately Lieer can't push emails back to GMail).

Or we could send the "bounce" messages using the regular MTA on the local machine. At least that would solve the problem of having to manually inspect the queue. But this solution risks make things a whole lot more complicated if one does not limit its scope.

flokli:
I meant to say, this needs to some big disclaimer explaining possible pitfalls. If people use this, they should be aware mails might not be sent at all.

I worry mostly about the feedback channel.

Cron might send emails if the job fails (assuming local mail delivery is configured properly) - but I wouldn't bet on every lieer user to have this configured properly.
Also, people might use systemd.timer in their user session - which also don't send emails on failures.

In addition to some scary docs, what about having the `queue` command itself checking if mails are stuck in the queue for too long, and then warn if you want to add more? I'm however not sure if MUAs properly show this output to the user…

gauteh:
Closing, see discussion.

skangas:
> Closing, see discussion.

What about it? The conclusion I draw is that the caveats need to be documented.

skangas:
Thinking about this a bit more, I'm not even sure it even needs that. stdout is sufficient error reporting for any command line program. In fact it is the only error reporting mechanism we have even without a queue.

(And if crontab and/or systemd is misconfigured and doesn't send emails on output, it is surely the responsibility of the user.)

gauteh:
Hm, maybe I misremember. I'll open again. I think my thought was that this should be kept fairly separate from the core gmi functionality. Maybe as a different package.

skangas:
> Hm, maybe I misremember. I'll open again. I think my thought was that this should be kept fairly separate from the core gmi functionality. Maybe as a different package.

It makes sense. We could make it into a new command "gmiq" if that is okay with you. I didn't yet find time to look into it, but it's in my backlog.

gauteh:
Stefan Kangas writes on July 7, 2020 11:52:
>> Hm, maybe I misremember. I'll open again. I think my thought was that this should be kept fairly separate from the core gmi functionality. Maybe as a different package.
>
> It makes sense. We could make it into a new command "gmiq" if that is okay with you. I didn't yet find time to look into it, but it's in my backlog.

Yes. That sounds good.

gauteh:
Closing, please reopen if new activity.

PR #148, opened 2020-05-12
add option for removing messages locally that have been deleted remot…

### LINKED ISSUE #125: option to preserve local copies of remotely-deleted messages
I'd like an option to tell lieer to keep local copies of any messages downloaded, even if they've been deleted remotely on gmail.

There are four reasons remote messages might have been deleted:

1. I did it because my quota is nearly full there
2. Someone broke into my gmail
3. I want to avoid google or legal authorities seeing them
4. They were in the spam folder and aged out of the 30-day spam retention window.

In the first three of these cases, I'd want to keep the local copies on my own machine.

Since spam is an exception, maybe there could be a way of applying to only not-in-spam-folder messages, or of doing a manual reap of local copies of deleted-on-remote messages which meet some criterion (like being between 30-40 days old and being marked spam), or some such.

Right now I get this functionality by periodically making a snapshot of the downloaded email, taking care to preserve sharing. But that's really not a very good solution.

gauteh:
…ely (fix #125)

by default set to true, this was the previous behavior when doing a
partial pull. but not when doing a full pull. when set to false and push
tries to update, this will fail with a message about 'could not find
remote message' which is otherwise ignored.

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -69,9 +69,6 @@ class Gmailieer:
     parser_pull.add_argument ('-f', '--force', action = 'store_true',
         default = False, help = 'Force a full synchronization to be performed')
 
-    parser_pull.add_argument ('-r', '--remove', action = 'store_true',
-        default = False, help = 'Remove files locally when they have been deleted remotely (forces full sync)')
-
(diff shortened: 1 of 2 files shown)


gauteh:
Merged.

Issue #139, opened 2020-05-05
gmi send croaks on some sendmail arguments


skangas:
I'm running the current master branch, where I'm very happy to see that there is now support for sending emails.

With the following configuration in GNU Emacs:

```
  (setq sendmail-program (expand-file-name "~/src/lieer/gmi"))
  (setq message-sendmail-extra-arguments '("send" "-C" "~/.mail/account.gmail"))
```

I type M-x message-mail in Emacs, but upon sending the email, I get these errors:

```
usage: gmi [-h] [--auth_host_name AUTH_HOST_NAME] [--noauth_local_webserver]
           [--auth_host_port [AUTH_HOST_PORT [AUTH_HOST_PORT ...]]]
           [--logging_level {DEBUG,INFO,WARNING,ERROR,CRITICAL}]
           {pull,push,send,sync,auth,init,set} ...
gmi: error: unrecognized arguments: -oi -f -t
```

I think Emacs is trying to pass some sendmail arguments, but lieer does not recognize them and croaks.  Maybe these arguments should be simply ignored?  I'm not sure.

Thanks in advance.

skangas:
This might be relevant, quoted from the exim sendmail man page:

```
       -f <address>
                 This option sets the address of the envelope sender of a  lo‐
                 cally-generated  message (also known as the return path).

       -i        This option, which has the same effect as -oi, specifies that
                 a dot on a line by itself should not terminate  an  incoming,
                 non-SMTP message.

       -t        When Exim is receiving a locally-generated, non-SMTP  message
                 on its standard input, the -t option causes the recipients of
                 the message to be obtained from the To:, Cc:, and Bcc: header
                 lines  in  the message instead of from the command arguments.
                 The addresses are extracted before any rewriting takes  place
                 and the Bcc: header line, if present, is then removed.
```

gauteh:
Yes, msmtp does the same. If any of those can be disabled in emacs config
that would be better, unless they are completely safe to ignore.

Do you know what those options are?

tir. 5. mai 2020, 16:57 skrev Stefan Kangas <notifications@github.com>:

skangas:
Thanks for your prompt reply here, I really appreciate it.

> Yes, msmtp does the same. If any of those can be disabled in emacs config that would be better, unless they are completely safe to ignore. Do you know what those options are?

My old configuration for msmtp was simply:

`
(setq sendmail-program "/usr/bin/msmtp")
`

[...]

PR #154, opened 2020-05-14
gmailieer: behave more like sendmail

### LINKED ISSUE #152: gmi send doesn't accept additional arguments

flokli:
sendmail(1) allows additional recipients to be passed in the command
line. The path to the message is never passed, it's always stdin:

> sendmail [option ...] [recipient ...]

At least neomutt makes use of the `recipient` arguments (and doesn't set
-t), so lieer needs to handle this properly.

Most sendmail implementations allow passing additional recipients in additional
arguments, but there's no way to express this to the GMail API - it basically
only supports the `-t` (`--read-recipients`) mode of sendmail, without
additional recipients.

However, MUA can make use of both `-t` and additional recipients, even in a
redundant fashion (with `-t` set, providing recipients in the message AND as
recipient in the CLI).

We need to check if the MUA asks us to do something we can't do.

The following combinations are OK:

 - When `-t` is passed, we need to check for the CLI-passed recipients to be
   equal or a subset of the ones passed in the headers.

 - When `-t` is not passed, all header-passed recipients need to be provided in
   the CLI as well.

This avoids silently not sending mail to some recipients (pretending we did),
or sending mail to recipients we didn't want to send to again.

Fixes #152

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -106,12 +106,13 @@ class Gmailieer:
       sys.argv.remove('-i')
 
     parser_send.add_argument('-i', action='store_true', default = None, help = 'Ignored: always implied, allowed for sendmail compatability.', dest = 'i3')
-    parser_send.add_argument('-t', action='store_true', default = None, help = 'Ignored: always implied, allowed for sendmail compatability.', dest = 'i0')
+    parser_send.add_argument('-t', '--read-recipients', action='store_true',
+                             default = False, dest = 'read_recipients')
 
     parser_send.add_argument('-f', type = str, help = 'Ignored: has no effect, allowed for sendmail compatability.', dest = 'i1')
 
-    parser_send.add_argument('message', nargs = '?', default = '-',
-        help = 'MIME message to send (or stdin "-", default)')
+    parser_send.add_argument('recipients', nargs = '*', default = [],
+        help = 'Recipients to send this message to')
 
     parser_send.set_defaults (func = self.send)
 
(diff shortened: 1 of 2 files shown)


lieer/gmailieer.py
```diff
@@ -707,21 +708,38 @@ def send (self, args):
...
+    header_recipients = set()
+    for field_name in ("To", "Cc", "Bcc"):
+      field_values = eml.get_all(field_name, [])
+      field_addrs = map(lambda x: email.utils.parseaddr(x)[1], field_values)
+      header_recipients = header_recipients.union(field_addrs)
+
+    if args.read_recipients:
+      if not header_recipients.issuperset(cli_recipients):
+          print ("Recipients passed via CLI, but not part of message headers: {}".format(", ".join(cli_recipients.difference(header_recipients))))
+          exit (2)
```
gauteh, line 736: Maybe use an exception here, just to stay consistent with rest of application.

lieer/gmailieer.py
```diff
@@ -707,21 +708,38 @@ def send (self, args):
...
+    elif not header_recipients == cli_recipients:
+      print ("Recipients passed via CLI differ from those in message headers:")
+      print (" - {}".format(", ".join(cli_recipients)))
+      print (" - {}".format(", ".join(header_recipients)))
+      exit (2)
```
gauteh, line 741: Same here.

lieer/gmailieer.py
```diff
@@ -707,21 +708,38 @@ def send (self, args):
...
-      msg = sys.stdin.buffer.read()
-      fn = 'stdin'
-    else:
-      if os.path.isabs(args.message):
-        fn = args.message
-      else:
-        fn = os.path.join(self.cwd, args.message)
-      msg = open(fn, 'rb').read()
+    msg = sys.stdin.buffer.read()
+    fn = 'stdin'
```
gauteh, line 712: We don't need this variable anymore, it will always be stdin.

gauteh:
Hi, I think this looks good, just a few comments. It would be good to get this tested wih emacs as well (@skangas, #139). I think that maybe the error messages will be hard to understand without context, or at least figure out how to fix them. Maybe link to the readme or some wiki?

lieer/gmailieer.py
flokli: Indeed. dropped `fn` here, and updated the `self.vprint` further down to not print `fn` anymore.

lieer/gmailieer.py
flokli: I now `raise argparse.ArgumentError` with the (condensed) error message.

lieer/gmailieer.py
flokli: I now `raise argparse.ArgumentError` with the error message.

flokli:
@gauteh I updated the error messages a bit, they now say
> "Recipients passed via sendmail(1) arguments" … vs "those in message headers"

IMHO, this should be clear enough - we can't really document how every MUA uses sendmail, and it can be altered through configuration anyways. `lieer` now just works in more of these possible combinations, to the extend of what the GMail API allows us to do without using SMTP.

Regarding emacs: My guess would be it just passed `-t`, didn't specify any additional recipients, and passed the message as stdin - otherwise I don't see how code in master could have worked at all. 

But I agree someone running an emacs setup should at least verify sending mail still works after this change.

gauteh:
Thanks, sounds good, but I can't see any changes? Did you push?

flokli:
:facepalm: now I did!

gauteh:
Thanks, merged, I changed ArgumentError to ValueError since the former needed
more arguments. I tweaked the readme a bit, trying to condense it a bit
more. It is already super-long. I think this was a very necessary
change, thanks.

skangas:
> Hi, I think this looks good, just a few comments. It would be good to get this tested wih emacs as well (@skangas, #139). I think that maybe the error messages will be hard to understand without context, or at least figure out how to fix them. Maybe link to the readme or some wiki?

I didn't have time to test it in time before it was merged. I will report back if I see any issues.

gauteh:
Stefan Kangas writes on May 19, 2020 2:18:
>> Hi, I think this looks good, just a few comments. It would be good to get this tested wih emacs as well (@skangas, #139). I think that maybe the error messages will be hard to understand without context, or at least figure out how to fix them. Maybe link to the readme or some wiki?
>
> I didn't have time to test it in time before it was merged. I will report back if I see any issues.

Great, the safest is to add the `-t` option in emacs as well. I
updated the wiki.

PR #155, opened 2020-05-30
Don't print messages about ignoring files outside our working path

### LINKED ISSUE #97: multiple email accounts in one notmuch database and file is not in this repository message
I do use single notmuch database which index emails from multiple accounts. These accounts end up having emails from different mailing list. That means i do have emails with same message-id duplicated in different filesystem directories. when running gmi sync I get the below messages always 

'/dir1/Inbox/cur/1532493124.4677_62.skywalker,U=514:2,ST' is not in this repository, ignoring.

dir1 where is I sync other email accounts and not gmail 
Following that, I always find that we are receiving metadata of about 47000 all the time. 

Anything I can do to avoid that 4k7 metadata update all the time.

-aneesh

georgyo:
If you are syncing two mailboxes to the same notmuch database, and those two mailboxes are both subscribed to an identical mailing list; you will end up with many many messages complaining that the message path isn't in the correct spot.

IE, messages like this will always have one side complain about the other side.
```
% notmuch show --format=json thread:0000000000000017 | jq '.[][][0] | {id: .id, filename: .filename, subject: .headers.Subject}'
{
  "id": "CAE16B76QGuYQ60BHH15f6tzTWdi-VXNvsQfjRyGeZkJZ53rZUg@mail.gmail.com",
  "filename": [
    "/home/shammas/Maildir/georgyo@gmail.com/mail/cur/17253b08bdd09ab7:2,",
    "/home/shammas/Maildir/george@shamm.as/mail/cur/17253b08b38e9a5a:2,"
  ],
  "subject": "[NYCResistor] May 26th 2020"
}
```

As this message is stating operating correctly, I think it is safe to remove the print.

fixes #97

### DIFF
--- a/lieer/local.py
+++ b/lieer/local.py
@@ -404,9 +404,7 @@ class Local:
 
     for m in msgs:
       for fname in m.get_filenames ():
-        if not self.contains (fname):
-          print ("'%s' is not in this repository, ignoring." % fname)
-        else:
+        if self.contains (fname):
           # get gmail id
           gid = self.__filename_to_gid__ (os.path.basename (fname))
           if gid:


gauteh:
Looks good. I agree.

PR #168, opened 2020-07-22
fix header_recipients calculation with mulitple addresses

flokli:
email.utils.parseaddr() is useful to parse a single email address into
its realname and mailaddress parts.

If there's multiple of these in the same line, email_utils.getaddresses
needs to be used for this to work.

This can be taken like that from
https://docs.python.org/3.8/library/email.utils.html.


gauteh:
Thanks, merged.

flokli:
Thanks for the quick merge! Can you craft a release fixing this?

PR #176, opened 2020-09-02
Add notmuch to setup.py

nackjicholson:
#174 Related

Need to install `notmuch` when running `python setup.py install`


gauteh:
Merged.

PR #179, opened 2020-10-08
notmuch: use notmuch2 bindings

### LINKED ISSUE #173: Migrate to notmuch2
I saw this announcement https://notmuchmail.org/news/release-0.30/ in which they say they've officially deprecated the `notmuch` bindings in favor of `notmuch2`.

I'm interested what you think might be involved in migrating the bindings to notmuch2? It could be a good task for anyone looking to contribute to the project.

An aside:
Why isn't `notmuch` in the requirements.txt? I've had to install it with `pip install notmuch` to get lieer to run.

gauteh:
missing `get_directory` in bindings

fixes #173.

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -23,7 +23,7 @@ import  argparse
 from    oauth2client import tools
 import  googleapiclient
 import  googleapiclient.errors
-import  notmuch
+import  notmuch2
 
 from .remote import *
 from .local  import *
@@ -312,8 +312,8 @@ class Gmailieer:
       self.remote.get_labels ()
 
     # loading local changes
-    with notmuch.Database () as db:
-      (rev, _) = db.get_revision ()
+    with notmuch2.Database() as db:
+      (rev, _) = db.revision ()
 
       if rev == self.local.state.lastmod:
         self.vprint ("push: everything is up-to-date.")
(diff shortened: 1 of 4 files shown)


sten0:
Wow, this looks much less painful than I thought that it would be!  P.S. I can confirm that Debian has a system-level `python3-notmuch2` package, so users of the Debian package (system-level) of Lieer should be able to use future versions of Lieer based on this PR, without the need for legacy compat.

Thank you for your continued work on Lieer, it has become essential software for me :-D

mjg:
I'm not sure whether the Ubuntu used in the test runner packages notmuch2 bindings, but apparantly you use the Ubuntu package for the lib but pull the notmuch2 package via pip, and neither provide the compiled cffi extension module (???) so that that pip tries to compile it - but misses a header file (which is in notmuch-devel or such).

I'd try this branch locally with my working notmuch2 bindings, but just today lieer is stuck in repeated "pull: historyId is too old, full sync required." (with both historyId and revision the same in the state file and the bak and the output), so something else needs fixing first before trying lieer with nm2.

sten0:
+1 for closing, belated thanks for this work, and it looks like Debian is
up-to-date: Debian
https://salsa.debian.org/debian/lieer/-/commit/72c298abcd794f150bac6ec9614c231b0a153295

PR #183, opened 2020-11-15
Update README to include alternate install method

ImranVirani:


gauteh:
I think `python setup.py install` should be replaced with `pip install .`, otherwise I think this looks good.

ImranVirani:
I've updated the file to include your suggestion.

gauteh:
Thanks, merged.

PR #190, opened 2021-03-02
Add --local-trash-tag option.

mkleehammer:
Here is a patch to add a --local-trash-tag option which updates the Local translation map.  It is minor and I've added it to the documentation.  Thoughts?


gauteh:
Hm, I think this looks pretty much good to go. Will test a bit.

tir. 2. mar. 2021, 20:50 skrev Michael Kleehammer <notifications@github.com

PR #191, opened 2021-03-15
update backoff recovery to be more aggressive

bshanks:
Partially recover from exponential backoff after every successful request/batch,
rather than waiting for 10 successes.

This appears to speed up large syncs. In large syncs, the current code (before this change is applied) appears to spend most of its time waiting about 31 seconds in between batches. This is unnecessary; as you can see by running with the proposed change, Google usually accepts a partial recovery from backoff after waiting 31 seconds.

As far as i can tell, https://developers.google.com/gmail/api/guides/handle-errors#exponential-backoff does not mandate waiting for multiple successes before recovering from backoff.

### DIFF
--- a/lieer/remote.py
+++ b/lieer/remote.py
@@ -333,16 +333,16 @@ class Remote:
       try:
         batch.execute (http = self.http)
 
-        # gradually reduce user delay if we had 10 ok batches
+        # gradually reduce user delay upon every ok batch
         user_rate_ok += 1
-        if user_rate_delay > 0 and user_rate_ok > 10:
+        if user_rate_delay > 0 and user_rate_ok > 0:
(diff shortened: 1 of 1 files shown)


gauteh:
Thanks, could this result in back off flipping between acceptable value and lower, rejected value? We could try it. 10 was definitely too high.

bshanks:
When I ran it, it appeared to occasionally flip but more often not, and overall, it appeared to result in an speedup compared to higher values (for example, stepping down after 2 successes seemed to me to be slower than stepping down after 1 success, which is why this PR steps down after 1 success). However, I didn't test it very rigorously.

gauteh:
Ok, thanks, merging.

PR #192, opened 2021-03-15
fix bug where saved local-trash-tag config option was ignored

bshanks:
My local-trash-tag config option was being ignored, even though "gmi set" showed:
```
Trash tag (local) .........: deleted
```

I think what was happening may have been that ```Local.update_translation('TRASH', self.local_trash_tag)``` was only being called from within ```set_local_trash_tag```, which i assume is called 
when --local-trash-tag was provided on the commandline, but not when it is found in the saved configuration.

This change runs ```Local.update_translation('TRASH', self.config.local_trash_tag)``` from within ```load_repository```.

### DIFF
--- a/lieer/local.py
+++ b/lieer/local.py
@@ -295,6 +295,7 @@ class Local:
     self.state = Local.State (self.state_f, self.config)
 
     self.ignore_labels = self.ignore_labels | self.config.ignore_tags
+    Local.update_translation('TRASH', self.config.local_trash_tag)
 
     ## Check if we are in the notmuch db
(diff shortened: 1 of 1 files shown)


gauteh:
Thanks, i think you are right. @mkleehammer, have you noticed this?

gauteh:
Merged.

mkleehammer:
Oy - I hadn't noticed, but I'm sure I some emails I now need to go delete :).  Thanks for finding and fixing this.  I'll pull the update now.

PR #193, opened 2021-03-15
add --translation-list-overlay option

bshanks:
This PR adds a --translation-list-overlay option that allows additions and changes to the label/tag translation list. For example,  

```
--translation-list-overlay CATEGORY_FORUMS,my_forum_tag,another_gmail_label,my_local_label
```

creates the mappings:
```
CATEGORY_FORUMS     <--> my_forum_tag
another_gmail_label <--> another_local_tag
```

To give some idea why this might be useful, I have two Gmail accounts and I use filtering on the server to forward certain emails from account1 to account2, and then delete them from account1. However, if both are being sync'd to my local machine, then this filtering rule assigns a 'TRASH' label to the email in account1, which then in some cases is propagated to account2, causing the email to be deleted everywhere, which is not what i want. I am planning to try mapping 'TRASH' on account1 to 'trash_account1', and similarly for INBOX and for account2. In this way, the inbox/trash state of the email will be distinct for my two accounts, even though notmuch only maintains one set of tags for each message-id.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -180,6 +180,11 @@ Lieer can be configured using `gmi set`. Use without any options to get a list o
 
   *Important*: See note below on [changing this setting after initial sync](#changing-ignored-tags-and-translation-after-initial-sync).
 
+**`Translation List Overlay`** can be used to add or change entries in the translation mapping between local and remote tags. Argument is a comment-separated list with an even number of items. This is interpreted as a list of pairs of (remote, local), where each pair is added to the tag translation overwriting any existing translation for that tag if any. For example,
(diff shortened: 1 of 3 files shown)


gauteh:
It would probably be better if both this and the --local-trash-tag functionality operated on the instance of `Local`, it is not inconceivable that lieer will be used as a library. And as far as I can tell there is not anything stopping it from that yet (except it would not be thread-safe to operate on the same repo from several instances).

bshanks:
I rebased on master, and changed both this and the --local-trash-tag functionality to operate on the instance of Local. I made translate_labels and labels_translate into instance variables, and renamed the class variables (which serve as hardcoded defaults) to translate_labels_default and labels_translate_default.

gauteh:
Thanks, merged, and added a simple test -- our first one!

PR #212, opened 2022-02-10
Notmuch2 interface updates

aplund:
I've tried to merge across the interface calls to notmuch2.

I have done some simple testing, but might have missed some things that have changed.

### DIFF
--- a/lieer/local.py
+++ b/lieer/local.py
@@ -22,7 +22,7 @@ import configparser
 from pathlib import Path
 import tempfile
 
-import notmuch
+import notmuch2
 from .remote import Remote
 
 class Local:
@@ -326,9 +326,9 @@ class Local:
       raise Local.RepositoryException ('local repository not initialized: could not find mail dir structure')
 
     ## Check if we are in the notmuch db
-    with notmuch.Database () as db:
+    with notmuch2.Database() as db:
       try:
-        self.nm_dir  = db.get_directory (os.path.abspath(self.md))
+        self.nm_dir  = db.get_directory(os.path.abspath(self.md))
         if self.nm_dir is not None:
           self.nm_dir = self.nm_dir.path
         else:
@@ -337,7 +337,7 @@ class Local:
 
         self.nm_relative = self.nm_dir[len(db.get_path ())+1:]
 
-      except notmuch.errors.FileError:
+      except notmuch2.errors.FileError:
         raise Local.RepositoryException ("local mail repository not in notmuch db")
 
     ## Lock repository
@@ -632,18 +632,15 @@ class Local:
       else:
         print ("(dry-run) tried to update tags on non-existant file: %s" % fname)
 
-    nmsg  = db.find_message_by_filename (fname)
+    nmsg = db.get(fname)
 
     if nmsg is None:
       if self.dry_run:
         print ("(dry-run) adding message: %s: %s, with tags: %s" % (gid, fname, str(labels)))
       else:
         try:
-          if hasattr (notmuch.Database, 'index_file'):
-            (nmsg, _) = db.index_file (fname, True)
-          else:
-            (nmsg, _) = db.add_message (fname, True)
-        except notmuch.errors.FileNotEmailError:
+          (nmsg, _) = db.add_message (fname, sync_flags = True)
+        except notmuch2.errors.FileNotEmailError:
           print('%s is not an email' % fname)
           return True
         nmsg.freeze ()
--- a/lieer/local.py
+++ b/lieer/local.py
@@ -326,19 +326,12 @@ class Local:
       raise Local.RepositoryException ('local repository not initialized: could not find mail dir structure')
 
     ## Check if we are in the notmuch db
-    with notmuch2.Database() as db:
+    with notmuch2.Database () as db:
       try:
-        self.nm_dir  = db.get_directory(os.path.abspath(self.md))
-        if self.nm_dir is not None:
-          self.nm_dir = self.nm_dir.path
-        else:
-          # probably empty dir
-          self.nm_dir = os.path.abspath (self.md)
-
-        self.nm_relative = self.nm_dir[len(db.get_path ())+1:]
-
-      except notmuch2.errors.FileError:
+        self.nm_relative=str(Path(self.md).relative_to(db.path))
+      except ValueError:
         raise Local.RepositoryException ("local mail repository not in notmuch db")
+      self.nm_dir=str(Path(self.md).resolve())
 
     ## Lock repository
     try:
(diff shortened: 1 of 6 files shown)


lieer/local.py
```diff
@@ -359,15 +352,9 @@ def load_repository (self, block = False):
 
     self.__load_cache__ ()
 
-    # load notmuch config
-    cfg = os.environ.get('NOTMUCH_CONFIG', os.path.expanduser('~/.notmuch-config'))
```
gauteh, line 363: Do we still support other configuration locations? Is this handled by notmuch?

lieer/local.py
```diff
@@ -632,52 +622,40 @@ def update_tags (self, m, fname, db):
...
-        # adding initial tags
-        for t in labels:
-          nmsg.add_tag (t, True)
-
-        for t in self.new_tags:
-          nmsg.add_tag (t, True)
+        with nmsg.frozen():
+          # adding initial tags
+          nmsg.tags |= labels
+          nmsg.tags |= self.new_tags
```
gauteh, line 642: do we not need tags to maildir flag sync anymore?

lieer/local.py
aplund: Notmuch now has a more sophisticated configuration handling mechanism.  It allows for multiple profiles and XDG standardised locations.  It would seem much better to allow notmuch to handle this as the user is very likely to have this set up already and it wouldn't make sense to re-implement it here.

lieer/local.py
aplund: I could not find any function for this in notmuch2 python interfaces.  It was my assumption that after the tags are changed, and if the notmuch config has 'synchronize_flags=true', then this would happen automatically.  Though, I'm not very familiar with what is going on behind the scenes here.

lieer/local.py
gauteh: Ok, are you sure this is is handled automatically with this method?

gauteh: What happens if a user has multiple notmuch databases, will notmuch figure out which config is present dependent on the dir that lieer is run from? or is it handled by env variables like before but notmuch does it?

aplund: It would seem to be dependant on the environment.  Presumably, as per what the code does here, other users would have to set the environment if they had multiple databases.  Therefore, I'm not sure if this change harms backwards compatibility.

The function calls into the notmuch library which does the configuration lookup.  The lookup is performed as per the manpage:
```
   CONFIGURATION
       Notmuch configuration file search order:

       1. File specified by notmuch --config global option; see notmuch(1).

       2. File specified by NOTMUCH_CONFIG environment variable.

       3. $XDG_CONFIG_HOME/notmuch/<profile>/config where <profile> is defined
          by   NOTMUCH_PROFILE   environment   variable   if   set,  $XDG_CON‐
          FIG_HOME/notmuch/default/config otherwise.

       4. $HOME/.notmuch-config.<profile> where <profile> is defined  by  NOT‐
          MUCH_PROFILE environment variable if set, $HOME/.notmuch-config oth‐
          erwise.
```

Maybe in the future the lieer configuration could specify a notmuch profile to use, if the user doesn't want the default.

lieer/local.py
aplund: I missed this.  So I checked the new API and you need to call `nmsg.tags.to_maildir_flags()`.  The call is within a method of the `MutableTagSet` instance.  I can update the PR (if I understand how github PRs go).

gauteh: Thanks, pushing a new commit to the branch you made your PR from will add it to the PR.

lieer/local.py
gauteh: I think this is fine. This matches the current behavior plus new options by notmuch.

gauteh:
Tests are failing. Is the notmuch2 version on pip the latest one?

aplund:
The error message in the test output relates the notmuch.h library header files missing.\

```
       gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -fPIC -I../../lib -I/opt/hostedtoolcache/Python/3.9.10/x64/include/python3.9 -c build/temp.linux-x86_64-3.9/notmuch2._capi.c -o build/temp.linux-x86_64-3.9/build/temp.linux-x86_64-3.9/notmuch2._capi.o
      build/temp.linux-x86_64-3.9/notmuch2._capi.c:572:14: fatal error: notmuch.h: No such file or directory
        572 |     #include <notmuch.h>
            |              ^~~~~~~~~~~
      compilation terminated.
      error: command '/usr/bin/gcc' failed with exit code 1
```

lieer/local.py
gauteh: Looks good!

gauteh:
So we probably need to install libnotmuch-dev and also notmuch-python if it exists as a package. Otherwise it needs to be installed from source. The python bindings requires a fairly new notmuch version, newer than ubuntu 18 has. 

What would be the preferred way to set up gmi on say ubuntu 18 and 20 now? We can no longer use the pypi package. And requirements.txt is probably just going to cause confusion, we probably have to point it to the notmuch git repository (pip+git).

aplund:
If libnotmuch-dev was installed, then the pip package would probably build.

gauteh:
Yeah, but the pip package is outdated and not maintained by notmuch org :/ (as far as I can see).

rex1fernando:
It seems that notmuch has no plans to keep the pip package up-to-date:

https://notmuchmail.org/pipermail/notmuch/2020/029683.html

gauteh:
Just to be clear. I think this PR is good to go, and thanks for all the work. I am hesitating to merge because it is difficult to install the new notmuch bindings. I am currently on Ubuntu 18 and there is no way to do that without compiling entire notmuch from source.

mjg:
Given the dependencies, I'd say notmuch and its bindings is something you want from your distro, not from pip. Ubuntu 18 LTS is based on Debian buster, and buster-backports has python3-notmuch2 at 0.31.4-2. I'm not a Debian nor Ubuntu guy and therefore don't know whether this helps.

0.31 is quite an early version of the notmuch2 bindings, though: Since then there were a couple of fixes and additions, especially around "split config".

Debian bullseye backports has 0.36 already (and Ubuntu 20 LTS is based on bullseye). I'm note sure how far you want to go back with LTS support, but the new notmuch2 bindings don't seem to be a good match for an 18 LTS when both 20 LTS and 22 LTS are around.

aplund:
I have tried to rewrite this to be as "dual stack" as possible.  I'm using a test on the module __name__ attribute to see which interface is being called and adapt.

It's very likely I've missed some of the interfaces so I'd appreciate help in testing.  I've tried running in a virualenv with either notmuch or notmuch2 installed and I've been able to sync mail with both.

I've also tried to fix the bug reported in issue #173 where the notmuch2 "set like" interface for the tags has seeminly changed.

mjg:
@aplund Thanks a lot for working on this.
@gauteh Do you really want to keep supporting the old bindings?
If yes: There is a lot of duplication in the compat code. Maybe we should go for a compatibility module which wraps these in one place instead?

gauteh:
No, I don't need to support the old bindings. Are the new bindings easy to install?

mjg:
> No, I don't need to support the old bindings. Are the new bindings easy to install?

There are those who prefer distro packages and those who prefer pip and such. As the maintainer of notmuch in Fedora, I may be biased :-)
I don't think there are official notmuch pip packages, and you want the bindings to match your notmuch. So the answer depends on the distro. Ubuntu 18LTS or so might be a problem. Fedora is fine, of course!

gauteh:
Sounds good. Then they're at least as well distributed as the old ones. I'm ready to merge this, but I can wait if @aplund or anyone wants to remove the old bindings first.

aplund:
Tried to remove old bindings.  Again, only done rudimentary testing.

gauteh:
Thanks, merged!

Issue #215, opened 2022-05-04
Migrate to non oob oauth before October


gauteh:
Will be blocked on October 3.

tohojo:
So any plans for this? October 3rd is less than a week away now :)

gauteh:
No, I don't have capacity atm, here are some links I got:

* https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html
* https://developers.google.com/identity/protocols/oauth2/resources/oob-migration

tohojo:
Alright, thanks for the pointers. I guess I'll take a look. It seems that the oauth2client library that lieer is using is deprecated in its entirety; do you have any opinions on whether it's more feasible to fix up the auth flow without replacing the library, or if it's better to just switch to one of the suggested alternatives while we're fixing up things? :)

gauteh:
Definitely better to switch to something that is maintained :)

-- gaute

mjg:
The funny thing is that they tell you to migrate to "Loopback IP address", which in turn is deprecated for most platforms (except desktop). I was hoping for a more permanent looking alternative to come up before next week ...
Note that oob will continue to work for apps in "test status", so if all else fails one should be able to use unchanged lieer with a personal api key (and weekly refresh, probably).

gauteh:
Not sure if we ever moved from test. Pretty sure it is no longer possible to create a similar key that public lieer has at the moment.

mjg:
https://github.com/getmail6/getmail6/blob/master/getmail-gmail-xoauth-tokens might be all that we need, but according to SO the expire times are 1h :|

tohojo:
I think it may be possible to refresh the token without user interaction, though, in which case that doesn't have to be such a big deal (just check and refresh it on each run). I'll experiment a bit with it and see how it works.

Side note, the documentation of that project (getmail6) recommends just using an app password instead of oauth; any reason lieer can't do that?

gauteh:
We already refresh the tokens, if you don't sync regularly you have to
reauth. Does app password work with the api? Also I think many orgs close
app passwords.

fre. 30. sep. 2022, 17:08 skrev Toke Høiland-Jørgensen <
***@***.***>:

tohojo:
> Does app password work with the api?

No idea; was hoping you knew ;)

[...]

PR #231, opened 2022-12-31
Use Loopback IP address auth flow instead of deprecated OOB

skangas:
This should fix the authorization flow, as discussed in #215. @tohojo already did all the heavy lifting and wrote up a patch, so I just tested that it works and made this pull request.

I would have actually preferred to put the commit in Toke's name, but I didn't want to do so without his consent. Instead, I acknowledge him in a `Co-authored-by` tag in the commit message. If Toke prefers, I'm obviously happy to make him the author instead. (My own changes were minimal.)

The patch works here, but more testing is obviously welcome.

For reference, see:
https://developers.googleblog.com/2022/02/making-oauth-flows-safer.html
https://developers.google.com/identity/protocols/oauth2/resources/oob-migration

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -145,6 +145,17 @@ class Gmailieer:
     parser_auth.add_argument ('-f', '--force', action = 'store_true',
         default = False, help = 'Re-authorize')
 
+    # These are taken from oauth2lib/tools.py for compatibility with its
+    # run_flow() method used during oauth
+    parser_auth.add_argument('--auth-host-name', default='localhost',
+        help='Hostname when running a local web server')
+    parser_auth.add_argument('--auth-host-port', default=[8080, 8090], type=int,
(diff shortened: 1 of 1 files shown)


gauteh:
Great! @tohojo do you wish to be `Author:` ?

tohojo:
I'm okay with co-authored-by; thanks for turning this into a proper PR! :)

gauteh:
Thanks, merged. Maybe instructions need to be updated as well? Does this close the auth issue?

skangas:
> Thanks, merged. Maybe instructions need to be updated as well? Does this close the auth issue?

I believe it does, or at least everything is now working on my end. I took a look at the README and everything there looks good to me from a cursory look.

gauteh:
Great!

lør. 31. des. 2022, 15:36 skrev Stefan Kangas ***@***.***>:

PR #245, opened 2023-05-31
add -v --verbose option for push & pull

digitalsignalperson:
from https://github.com/gauteh/lieer/issues/241 to show verbose logging the same as --dry-run


gauteh:
Thanks, seems like you need to add the flag to the mock-class as well.

PR #248, opened 2023-06-04
Add verbose option to sync

kode54:
Fixes f7c3f594ee4a04c8289e8431df9d2e5470affc56 which broke sync:

```
  File "/usr/bin/gmi", line 24, in <module>
    g.main ()
  File "/usr/lib/python3.11/site-packages/lieer/gmailieer.py", line 236, in main
    args.func (args)
  File "/usr/lib/python3.11/site-packages/lieer/gmailieer.py", line 307, in sync
    self.setup (args, args.dry_run, True)
  File "/usr/lib/python3.11/site-packages/lieer/gmailieer.py", line 281, in setup
    self.verbose          = args.verbose
                            ^^^^^^^^^^^^
AttributeError: 'Namespace' object has no attribute 'verbose'
```

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -135,6 +135,9 @@ class Gmailieer:
     parser_sync.add_argument ('-d', '--dry-run', action='store_true',
         default = False, help = 'do not make any changes')
 
+    parser_sync.add_argument ('-v', '--verbose', action='store_true',
+        default = False, help = 'print list of changes')
+
     parser_sync.add_argument ('-f', '--force', action = 'store_true',
         default = False, help = 'Push even when there has been remote changes, and force a full remote-to-local synchronization')
 
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -281,9 +281,13 @@ class Gmailieer:
         raise NotADirectoryError("error: %s is not a valid path!" % args.path)
 
     self.dry_run          = dry_run
-    self.verbose          = args.verbose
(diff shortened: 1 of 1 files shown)


kode54:
Oops, looks like the other modes also needed the exception fallback, because the base config model doesn't define the verbose flag.

gauteh:
Thank you, so this is not ready to merge yet?

@digitalsignalperson did you test the different gmi subcommands?

digitalsignalperson:
Oops admittedly I did not test sync with verbose mode, only push and pull. I'm only at the point of manually doing deliberate pushes and pulls and making sure I understand how everything changes on both sides (hence adding the verbose mode to help with this), so I haven't used the sync command ever yet! Sorry I guess I broke sync!

gauteh:
👍 Maybe we can find a way to do this in the unit tests eventually.

digitalsignalperson:
Hmm, I guess the intention is that `gmi pull --verbose` and `gmi sync --verbose` would both set `self.verbose = ` to True

kode54:
The exception I added in the second commit will set self.verbose to False for all modes which do not have the verbose option defined in their option model. Unless it would be preferred to move the verbose option to the common model? Does it even affect any of the other operations?

I was experiencing the exception error on the `gmi set` method as well, which was why I decided to add a generic catch-all.

gauteh:
It is better to add verbose to the common options than using exceptions for this.

jonrubens:
I get this error on `gmi init account`

gauteh:
Fixed through #252.

PR #252, opened 2023-07-17
Add --verbose flag for all commands.

sstepashka:
Recent commit ([add -v --verbose option for push & pull to show same changes as --dry…](https://github.com/gauteh/lieer/commit/f7c3f594ee4a04c8289e8431df9d2e5470affc56) added new `--verbose` flag for `pull` and `push` command, but didn't add it for `sync`. `setup` function checks the flag unconditionally what causes error for `sync` command. Adding `--verbose` for `sync` command solves the issue.

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -135,6 +135,9 @@ class Gmailieer:
     parser_sync.add_argument ('-d', '--dry-run', action='store_true',
         default = False, help = 'do not make any changes')
 
+    parser_sync.add_argument ('-v', '--verbose', action='store_true',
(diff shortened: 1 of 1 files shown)


gauteh:
Thank you very much. It would be a bit better to add the sync option to the common configuration options, do you have the chance to remove the specific ones and add it to the common one?

(https://github.com/gauteh/lieer/pull/248#issuecomment-1629179199)

sstepashka:
Hi @gauteh, I just did. PTAL

gauteh:
Thanks!

PR #253, opened 2023-09-01
Make GitHub workflow test also on Python 3.10-3.11

skangas:
This should make the GitHub test workflow test also with newer versions of Python.

More information:
https://docs.github.com/en/actions/automating-builds-and-tests/building-and-testing-python


gauteh:
Thanks 👍

PR #269, opened 2023-12-03
Enable more ruff rules

skangas:
This PR enables a bunch more ruff rules. They have quite a long list of various rules, so I picked some of the more low-hanging fruit here.

Some of the fixes are arguably mostly stylistic (e.g. `isort` might help when merging, but doesn't do much else), but some of them might improve performance (`Perflint`, `flake8-comprehensions`), even if only slightly.

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -862,8 +862,7 @@ class Gmailieer:
             self.bar.total = total
             self.bar_update(len(gids))
 
-            for m in gids:
-                message_gids.append(m["id"])
+            message_gids = [m["id"] for m in gids]
 
             if self.limit is not None and len(message_gids) >= self.limit:
                 break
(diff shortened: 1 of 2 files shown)


lieer/gmailieer.py
```diff
@@ -861,8 +862,7 @@ def full_pull(self):
             self.bar.total = total
             self.bar_update(len(gids))
 
-            for m in gids:
-                message_gids.append(m["id"])
+            message_gids = [m["id"] for m in gids]
```
gauteh, line 865: This always confuses me in python: doesn't this change the scope of message_gids?

skangas: Yeah, scoping in Python is weird and confusing. But since there is no block scoping in Python, only function scoping, this change does not change the scope of `message_gids`.

Compare to something like:

```
>>> def foo():
...     if True:
...         x = "wow"
...     print(x)
... 
>>> foo()
wow
```

gauteh: Ok, thanks, looks good.

PR #271, opened 2023-12-13
Fix my last change (Closes #270)

### LINKED ISSUE #270: gmi pull only pulls one mail.
`gmi pull` only pulls one mail with notmuch default settings while my gmail contains 1000+ mails.

skangas:
It seems like there was a mistake in #269. This should fix it, tested by pulling all of my emails in a fresh directory and then synching.


gauteh:
Thanks

PR #272, opened 2023-12-13
Enable ruff rules: executable, bandit

skangas:
This enables two more ruff rules: flake8-executable for consistency, and flake8-bandit for security checks.

I disabled the rule about asserts, which has been described as "alarmist at best" by the flake8 maintainer: https://stackoverflow.com/a/68429294


gauteh:
Thanks

PR #275, opened 2024-02-15
Require Notmuch 0.30+ for `notmuch2` Python bindings

tschwinge:
As of #173 "Migrate to notmuch2" etc., Lieer uses the `notmuch2` Python bindings, which have first been shipped in Notmuch 0.30.  (..., and 0.30 appears to still work fine for current Lieer, as I've just tried.)

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -20,7 +20,7 @@ While Lieer has been used to successfully synchronize millions of messages and t
 ## Requirements
 
 * Python 3
-* `notmuch >= 0.25` python bindings
+* Notmuch 0.30+ for `notmuch2` Python bindings
 * `google_api_python_client` (sometimes `google-api-python-client`)
 * `google_auth_oauthlib`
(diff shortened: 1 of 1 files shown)


gauteh:
Thanks, should probably be added to setup.py and requirements.txt

tschwinge:
So I looked into that -- and found out that <https://pypi.org/project/notmuch2/> is actually (a) not based on the upstream Notmuch sources, (b) unmaintained (last changed four years ago), and (c) only offers a `0.30-rc2` pre-release from Jun 19, 2020.  (I'm using my own local build of Notmuch.)

```
ERROR: Could not find a version that satisfies the requirement notmuch2>=0.30 (from versions: 0.1, 0.30rc2)
ERROR: No matching distribution found for notmuch2>=0.30
```

gauteh:
Oh right. Yeah, I remember that was a mess. Let's merge as is then.

PR #276, opened 2024-03-05
Handle the old python3-google-auth in Debian (1.5.x)

julian-klode:
Implement our own to_json() method and adjust the code handling the refresh of the credentials: Old versions of the library did not read expiry or token from the file, needing you to refresh the token, but we did not trigger refresh because the token was never considered expired, hence check whether we have a token, and , if not, also consider it expired.

Additionally, make the credential storage atomic such that we do not accidentally write an empty credential file on errors.

### DIFF
--- a/lieer/remote.py
+++ b/lieer/remote.py
@@ -15,6 +15,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program.  If not, see <https://www.gnu.org/licenses/>.
 
+import json
 import os
 import time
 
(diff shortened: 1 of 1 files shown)


julian-klode:
This is a bit sad, I have no idea why the library is stuck on 1.5.x in Debian, but I figured let's just make this work. This means doing a refresh on every sync which is annoying but that is the old library version for you.

gauteh:
Thanks. Finally merged this.

PR #281, opened 2024-06-15
Specifically mention git-send-email in setup

riking:
Looking up the git config key isn't the most obvious thing, so this will make it easier for people to migrate off app-specific passwords. There are _so many_ places online where that is listed as the only possible way to make git-send-email work.


gauteh:
Thanks, merged.

PR #282, opened 2024-07-08
Remove tests from installed packages

granquet:
The tests do not need to be installed

Create an exclude rule for find_packages() in setup

This is a slight issue that prevents packaging "cleanly" for gentoo, details here:
https://projects.gentoo.org/python/guide/qawarn.html#stray-top-level-files-in-site-packages


gauteh:
Thank you. Merged.

PR #287, opened 2024-09-07
Only update self.bar.total when self.bar is instantiated

gongzhitaao:
The `self.bar.total = total` is called even when `self.bar` is not instantiated.  This results in crash.


gauteh:
Thanks, merged.

PR #295, opened 2025-01-03
Update documentation on using your own OAuth 2 client

bcspragu:
Fixes https://github.com/gauteh/lieer/issues/292


gauteh:
Thanks!

gauteh:
The image links don't work in the readme, they need to be absolute links. The setup is like that to have the homepage which is necessary to get credentials from google.

PR #296, opened 2025-01-20
Fix links in documentation

bcspragu:
I didn't realized that the `README.md` was a symlink to `docs/index.md`. As a result, the images in the main README.md are broken, but looked fine when viewing `docs/index.md` directly. The solution is to use absolute paths.

Context: https://github.com/gauteh/lieer/pull/295#issuecomment-2600780145


gauteh:
Thanks a lot!

PR #321, opened 2026-04-20
local: gracefully fail if the lock is already taken

laarmen:
An EACCES or EAGAIN error on a non-blocking lockf is not actually an error, but rather the program working as intended. At this stage, any other error is unexpected and should probably be left alone.

The tailor-made exception can be caught in our main function to display a short error string and use a specific error code for the user to act accordingly. This is much more user-friendly (and machine-friendly) than the standard python uncaught exception handler.

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -416,7 +416,11 @@ class Gmailieer:
         if args.quiet:
             args.no_progress = True
 
-        args.func(args)
+        try:
+            args.func(args)
+        except Local.LockingException as e:
+            print(e, file=sys.stderr)
+            sys.exit(7)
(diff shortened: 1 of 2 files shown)


gauteh:
Hi, what's the point of this PR? It seems to just fail with a slightly different error message?

laarmen:
The point of this PR is to differentiate this error, which is a frequent and benign user error. Since it's benign, it shouldn't fill your terminal with a stack trace, and it's nice to have a specific error code (randomly, 7), so that any automation can just deal with it accordingly, e.g. simply suppress it.

gauteh:
Ok. Good, merging.

PR #322, opened 2026-04-20
send: add a non-blocking option

laarmen:
I personally prefer this operation to fail fast as sync operations sometimes take a *lot* of time.

### DIFF
--- a/lieer/gmailieer.py
+++ b/lieer/gmailieer.py
@@ -194,6 +194,13 @@ class Gmailieer:
         )
 
         parser_send.add_argument(
+            "--blocking",
+            action=argparse.BooleanOptionalAction,
+            default=True,
+            help="Whether the call should block on other operations on the local repository, or error out. Defaults to blocking.",
+        )
+
+        parser_send.add_argument(
             "-f",
             type=str,
             help="Ignored: has no effect, allowed for sendmail compatibility.",
@@ -1011,7 +1018,7 @@ class Gmailieer:
             return ResumePull.new(f, lastid)
 
     def send(self, args):
-        self.setup(args, args.dry_run, True, True)
+        self.setup(args, args.dry_run, True, args.blocking)
         self.remote.get_labels()
 
         msg = sys.stdin.buffer.read()


gauteh:
Hi. If the default is True, and the switch sets it to True, wont it always be True..? Have you tested this?

laarmen:
I actually added this for the `--no-blocking` option

gauteh:
> I actually added this for the `--no-blocking` option

The code says --blocking.

laarmen:
`BooleanOptionalAction` gives you both the positive flag and its negative counterpart. I can implement it manually if you prefer, but flags cancelling each other are pretty handy in my experience.
```
❯ gmi send --help
usage: gmi send [-h] [-C PATH] [-c CREDENTIALS] [-s] [-q] [-v] [-d] [-i] [-t] [--blocking | --no-blocking] [-f I1] [recipients ...]

Read a MIME message from STDIN and send.

positional arguments:
  recipients            Recipients to send this message to (these are essentially ignored, but they are validated against the header fields.)

options:
  -h, --help            show this help message and exit
  -C, --path PATH       path
  -c, --credentials CREDENTIALS
                        optional credentials file for google api
  -s, --no-progress     Disable progressbar (always off when output is not TTY)
  -q, --quiet           Produce less output (implies -s)
  -v, --verbose         print list of changes
  -d, --dry-run         do not actually send message
  -i                    Ignored: always implied, allowed for sendmail compatibility.
  -t, --read-recipients
                        Read recipients from message headers. This is always done by GMail. If this option is not specified, the same addresses (as those in the headers) must be specified as additional arguments.
  --blocking, --no-blocking
                        Whether the call should block on other operations on the local repository, or error out. Defaults to blocking.
  -f I1                 Ignored: has no effect, allowed for sendmail compatibility.
  
❯ echo foo | gmi send -C /Users/schopin/Mail/personal/ -t --no-blocking
path: /Users/schopin/Mail/personal/
failed to lock repository (probably in use by another gmi instance)
```

(Also using #321 in this case)

gauteh:
Ok, good, thanks for the explanation.

PR #323, opened 2026-04-20
local: add support for database.mail_root notmuch config option

laarmen:
This option was introduced in 0.32, and allows for the DB path to be decorrelated to the maildir location. See notmuch-config(1) for details.

### DIFF
--- a/lieer/local.py
+++ b/lieer/local.py
@@ -367,8 +367,9 @@ class Local:
 
         ## Check if we are in the notmuch db
         with notmuch2.Database() as db:
+            mail_root = db.config.get("database.mail_root", db.path)
             try:
-                self.nm_relative = str(Path(self.md).relative_to(db.path))
+                self.nm_relative = str(Path(self.md).relative_to(mail_root))
             except ValueError:
                 raise Local.RepositoryException(
                     "local mail repository not in notmuch db"


gauteh:
We still require it to be relative / sub-dir to the db path?

laarmen:
AFAICT we don't really care about the database path, it's the maildir that's important. The DB used to be located next to the maildir, but with that option they can now be de-correlated, which I believe is now the default behavior on new configurations, if I read this comment correctly:

```
# Supported values are 'mail_root' and 'path'. The recommended option
# is to set 'mail_root' to the directory where your mail currently exists
# and to where mail will be delivered in the future. Files should be
# individual email messages. By default notmuch will store its database
# in $XDG_DATA_HOME/notmuch; you can override this by setting 'path'.
# If only 'path' is set, this directory is for 'mail_root' and for
# the database location (in a subdirectory called ".notmuch").
#
[database]
mail_root=/Users/schopin/Mail
path=/Users/schopin/.local/share/notmuch/
```

gauteh:
Ok. Good. I don't have the chance to test this, but if you've had the chance to test it then I'm happy.

laarmen:
> Ok. Good. I don't have the chance to test this, but if you've had the chance to test it then I'm happy.

I've been using it for about a month, my setup doesn't work without this
patch.

gauteh:
Ok, thanks.

